PR #1, opened 2019-07-12
Add MVP cloud function to import histograms

robhudson:


functions/db/import_histograms.fn.js
```diff
@@ -0,0 +1,95 @@
+const admin = require('firebase-admin');
+const { BigQuery } = require('@google-cloud/bigquery');
+const functions = require('firebase-functions');
+const { murmurHash128 } = require('murmurhash-native');
```
openjck, line 6: If this project targets [a platform that supports `ECMAScript modules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Browser_compatibility), these could be rewritten to be more modern like so:

```javascript
import admin from 'firebase-admin';
import { BigQuery } from '@google-cloud/bigquery';
import functions from 'firebase-functions';
import { murmurHash128 } from 'murmurhash-native';
```

functions/db/import_histograms.fn.js
```diff
@@ -0,0 +1,95 @@
...
+
+try {
+  admin.initializeApp();
+} catch (e) {
+  console.log(e);
+}
```
openjck, line 12: It looks like the rest of the code in this file will run even if `admin.initializeApp()` throws an error. Would it make sense to put this `try`/`catch` block inside the export that begins on line 84?

functions/db/import_histograms.fn.js
```diff
@@ -0,0 +1,95 @@
...
+  const bigquery = new BigQuery();
+  const query = `
+    SELECT *
+    FROM \`moz-fx-data-derived-datasets.analysis.client_probe_counts\`
+    WHERE
+      channel = @channel
+      AND app_version = @app_version
+      AND metric = @metric
+      AND os = @os
+  `;
```
openjck: Nice use of template strings! 👍

functions/db/import_histograms.fn.js
```diff
@@ -0,0 +1,95 @@
...
+  const options = {
+    query,
```
openjck: You're an ES6 pro! 👍

functions/db/import_histograms.fn.js
```diff
@@ -0,0 +1,95 @@
...
+    location: 'US',
+    params: {
+      channel,
+      app_version: version,
+      metric: 'scalar_content_browser_usage_plugin_instantiated',
+      os: 'Windows_NT'
+    }
+  };
+  const [rows] = await bigquery.query(options);
```
openjck: Small thing, but in my experience, `const [ rows ]` is more conventional and may be easier to read by some people as a result. I did a double-take on this line.

functions/db/import_histograms.fn.js
```diff
@@ -0,0 +1,95 @@
...
+  const docs = await readAggregates('release', '65.0');
+  const promises = docs.map(doc =>
+    writeDoc(doc.collection, doc.docId, doc.doc)
+  );
+  const results = await Promise.all(promises);
+  return results;
+}
+
+// eslint-disable-next-line no-multi-assign
+exports = module.exports = functions.https.onRequest((request, response) => {
```
openjck, line 83: Like before, depending on the platform being targeted, this could be rewritten like:

```javascript
export default functions.https.onRequest((request, response) => {
```

At least I think that works. It may have to be something like this instead:

```javascript
export default () => {
  functions.https.onRequest((request, response) => {
  }
};
```

functions/db/import_histograms.fn.js
```diff
@@ -0,0 +1,95 @@
...
+  // TODO: Get latest version info and iterate over all combinations of the last
+  // 10 versions + 3 channels.
+  importRows()
+    .then(() => {
+      return response.status(200).send('Probes imported.');
+    })
+    .catch(error => {
+      console.error(error);
+      response.status(500).send(error.message);
```
openjck: For consistency, `return` could be used in both `.then()` and `.catch()` or in neither. I don't think I would `return` in either in this case, but maybe the Airbnb style guide has a preference.

functions/db/import_histograms.fn.js
```diff
@@ -0,0 +1,95 @@
...
+    .doc(docId)
+    .set(doc);
+}
+
+async function importRows() {
+  const docs = await readAggregates('release', '65.0');
+  const promises = docs.map(doc =>
+    writeDoc(doc.collection, doc.docId, doc.doc)
+  );
+  const results = await Promise.all(promises);
```
openjck: Is `await` needed here? It looks like the return value of `importRows` is given a callback on line 88.

openjck: Maybe it is. I'm a little confused because I think `async` functions like `importRows` are usually wrapped in a Promise themselves.

openjck (commented):
Made a few comments. :smiley:

functions/db/import_histograms.fn.js
robhudson: Cloud functions are node.js and it looks like it would require naming these as `.mjs` files.

functions/db/import_histograms.fn.js
robhudson: Good catch. I would want this cloud function to fail if there was an error here otherwise we probably couldn't talk to Firestore.

functions/db/import_histograms.fn.js
openjck: Ah yeah. That may be true. I think newer versions of Node do away with that requirement but I'm not certain.

functions/db/import_histograms.fn.js
openjck: Also, as we talked about in IRC, this just grabs the first element from the array that `bigquery.query()` returns and assigns it the name `rows`.

functions/db/import_histograms.fn.js
robhudson: Actually, the way I'm importing these methods from the index.js file requires this... the app may already be initialized and this just catches it if it is a 2nd initialization.

openjck:
Nice improvements! Looks good to me. :smiley:

PR #3, opened 2019-08-08
Add Python import to be run on airflow

robhudson:


docker/importHistograms.py
```diff
@@ -0,0 +1,136 @@
...
+
+        print("Done!")
+
+
+def get_major_version(v):
+    return int(v.split(".")[0])
+
+
+def run():
+    j = json.loads(urllib.request.urlopen(PRODUCT_DETAILS_URL).read())
```
emtwo: nit: a better variable name than `j`? Maybe `firefox_releases` or something

docker/importHistograms.py
```diff
@@ -0,0 +1,136 @@
...
+    versions = {
+        "release": get_major_version(j["LATEST_FIREFOX_VERSION"]),
+        "beta": get_major_version(j["LATEST_FIREFOX_DEVEL_VERSION"]),
+        "nightly": get_major_version(j["FIREFOX_NIGHTLY"]),
```
emtwo: Just a thought here. In the data, there were a ton of junk channels but I filtered it all out except for the top 6:
"release", "esr", "beta", "aurora", "default", "nightly"

I'm actually not sure what `default` is and whether or not we care about esr or aurora to show up in MDV2. But depending on the answers to those questions, this code here might be affected.

docker/importHistograms.py
```diff
@@ -0,0 +1,136 @@
...
+
+        if i % 500 == 0 and i > 0:
+            batch = commit_batch(batch)
+
+        doc = {
+            "metric": row.metric,
+            "build_id": row.app_build_id,
+            "os": row.os,
+            "agg_type": row.agg_type,
+            "aggregates": row.aggregates["key_value"],
```
emtwo: Just a heads up that `key_value` got removed from `main_summary` as an intermediate. So I tried to follow suit and remove it from `client_probe_counts_v2`. So I think once you use that one, you might just be able to use `row.aggregates` directly instead of `row.aggregates["key_value"]`.

emtwo (commented):
Super neat! Left just a couple of comments with thoughts as I was reading, but overall looks good

docker/importHistograms.py
robhudson: Ah, good catch. I didn't even think that I was still using the `v1` table.

docker/importHistograms.py
```diff
@@ -0,0 +1,136 @@
...
+
+    query = """
+        SELECT *
+        FROM `{table}`
+        WHERE
+            channel = @channel
+            AND app_version = @app_version
+            AND os IS NOT NULL
+            AND app_build_id IS NOT NULL
+    """.format(
```
robhudson: @emtwo Does this query seem appropriate, with the `IS NOT NULL` clauses?

emtwo: Ah good question, I missed this. The data purposely has rows where `os` and/or `app_build_id` is null because those include *all* values for them. So we probably still need that data if it's possible to have no value set for them in the UI.

robhudson: The prototype didn't allow for selecting the OS or build ID so it's probably best to pull the `NULL` rows in as well. If I understand correctly, that would only add 1 extra row for each per metric, correct?

emtwo: It will add 3 extra rows per metric:
* os=null & build_is is not
* build_id=null & os is not,
* both are null

robhudson:
A new approach using dataproc will replace what is here. Closing...

PR #4, opened 2019-08-29
Small svelte mdv2 prototype

hamilton:


src/App.svelte
```diff
@@ -1,37 +1,76 @@
...
+  <Form>
+    <FormGroup>
+      <!-- Lots of boiler plate here. Any way to make this better? -->
+      <Selector formId="metric" label={'Select Metric:'} current={$STORE.metric} onChange={metric => dispatch(changeMetric(metric))} options={options.metric} />
+      <Selector formId="channel" label={'Select Channel:'} current={$STORE.channel} onChange={channel => dispatch(changeChannel(channel))} options={options.channel} />
+      <Selector formId="version" label={'Select Version:'} current={$STORE.version} onChange={version => dispatch(changeVersion(version))} options={options.version} />
+    </FormGroup>
+  </Form>
+
+  <Chart />
```
hamilton, line 33: One thing that might be worth trying would be to make this `<Chart data={$DATA} />`. When `$DATA` updates, I believe it should reactively propagate down into the children components.

src/Chart.svelte
```diff
@@ -1,25 +1,9 @@
...
-            [
-                {
-                    type: 'scatter',
-                    x: data.aggregates.map(d => d.key),
-                    y: data.aggregates.map(d => d.value),
-                }
-            ]
-        );
-    }
+<div id="chart"></div>
```
hamilton: you can actually keep `drawChart` here if you do decide to pass down data. One way that you can achieve this would be to also have this be `<div bind:this={chartElement} />` and define `let chart;` above. Then reactively update the plotly plot with `$: drawChart(data)` or whatever.

hamilton: In this case, `drawChart` would have to make sure `chartElement` is cleared each time, but that should be easy to do. No need for the `document.getElementById` if you use `bind:this`.

src/store.js
```diff
@@ -0,0 +1,91 @@
...
+export default { subscribe: STORE.subscribe }
+
+
+
+function createDataStore() {
+    const { subscribe, set, update } = writable({});
+
+    return {
+        subscribe,
+        storeData: (data) => update(d => data),
```
hamilton: this could probably just be `storeData: update`, right?

src/App.svelte
```diff
@@ -1,37 +1,76 @@
...
+    let chartContainer = document.getElementById('chart');
+    if (!chartContainer) return;
+
+    // Clear out any old chart objects.
+    while (chartContainer.firstChild) {
+      chartContainer.removeChild(chartContainer.firstChild);
+    }
+
+    // Create new chart on the fly.
+    let chartObj = document.createElement('div');
```
hamilton: if you try `bind:this` as described below, you probably don't need to create a new element, right? Does `Plotly.plot` overwrite whatever is in `chartObj`? These are minor quibbles, just curious.

src/App.svelte
```diff
@@ -1,37 +1,76 @@
...
+      .then(snapshot => {
+        snapshot.forEach(doc => {
+          DATA.storeData(doc.data().aggregates);
         });
+      });
+    renderChart($DATA);
+    console.log($DATA);
+  }
+
+  $: queryData($serverParams);
```
hamilton: it might be worth trying a pattern where we don't have `queryData` reactively called from within a component, but as a derived store triggered by `$serverParams` updating that is imported into `App.svelte`. This would allow one to separate out more of the important app-level logic from the components. One thing I have learned is that `$:` can get awfully tricky to reason about when more code is built up around it, sometimes detrimentally. That's one of the reasons why I feel pretty strongly about trying to have only _stores_ imported into components, which should always reflect the app state generated by the main store + stores derived from it.

hamilton (commented):
Looks great @robhudson! Left a couple of comments based on some other patterns I've used in the past – might be worth seeing if they work out.

I should note that this exercise of talking about it has helped _me_ clarify my own mental models. It's easy to take for granted some of this domain experience I've formed from working on other Svelte-based products. 👍

src/App.svelte
hamilton: here's a minimal example of how that might work: https://svelte.dev/repl/197888e11ef7416c8b1cf57b6d119805?version=3.9.1

src/store.js
robhudson: Tried that and got an error...
`TypeError: fn is not a function`

PR #5, opened 2019-09-03
Update probes to include more information for front-end

robhudson:


functions/db/import_probes.fn.js
```diff
@@ -25,58 +56,88 @@ async function storeProbe(key, probe) {
...
+      getJSONContent(PROBES_URL)
+        .then((probes) => {
 
-      res.on('data', d => {
-        data += d;
-      });
+          const promises = [];
+          Object.keys(probes).forEach(k => {
+            const probe = probes[k];
+            const oldestHistory = (
```
hamilton: This looks like it's the newest version of the first channel this probe was on. Could you add a comment above this stating that so it's as bit clearer to future me's?

hamilton (commented):
Looks good to me! Made one comment re: naming, otherwise looks great. Feel free to merge at your convenience.

functions/db/import_probes.fn.js
robhudson: Yeah it seems backwards, but `history.nightly[0]` is the most recent set of data and `history.nightly.slice(-1)[0]` is the oldest. I was thinking for the `description`, `kind`, and a few other fields we should get the oldest. But now I'm thinking if they changed maybe we should get the newest values for those.

PR #7, opened 2019-09-06
Design Scaffold

hamilton:
This is a bit of scaffolding for the new design, which itself is a WIP. It adds:

- `src/components` – the rough design system components (layout sections, svelte components)
- `src/app` – the store + component composition ("the app")
- `stories` – a handful of storybooks for some of the different components


package.json
```diff
@@ -10,17 +13,21 @@
...
     "sirv-cli": "^0.4.4",
-    "svelte": "^3.0.0"
+    "svelte": "^3.0.0",
+    "svelte-loader": "^2.13.6"
   },
   "dependencies": {
+    "d3-format": "^1.4.1",
     "firebase": "^6.3.0",
+    "flexsearch": "^0.6.30",
     "plotly.js-basic-dist": "^1.48.3"
```
robhudson: We could take firebase and plotly out of here if you're not using them.

hamilton: good point

src/app/store/telemetry-search.js
```diff
@@ -0,0 +1,44 @@
+import { readable } from 'svelte/store';
+
+import FlexSearch from 'flexsearch'
+const url = 'https://probeinfo.telemetry.mozilla.org/firefox/all/main/all_probes'
```
hamilton, line 4: @robhudson this can switch to the API endpoint.

src/app/store/telemetry-search.js
```diff
@@ -0,0 +1,44 @@
...
+
+const resultToProps = (info, id) => {
```
hamilton, line 6: @robhudson this function can probably be removed entirely since your ingestion job does this

src/app/store/telemetry-search.js
```diff
@@ -0,0 +1,44 @@
...
+    const out = {
+        name: info.name, probeType: info.type, description, id, firstRelease, firstReleaseName
+    }
+    return out;
+}
+
+const telemetrySearch = readable({loaded: false}, async (set) => {
+    let data = await fetch(url).then(r=>r.json())
+    data = Object.values(data).map(resultToProps)
+    const search = new FlexSearch({
```
hamilton, line 29: This is the `FlexSearch` instance that we'll need to tweak to perform better with telemetry probes (mostly, we'll need to add punctuation and fuzz a bit more)

package.json
```diff
@@ -10,17 +13,21 @@
     "rollup-plugin-svelte": "^5.0.3",
     "rollup-plugin-terser": "^4.0.4",
     "sirv-cli": "^0.4.4",
-    "svelte": "^3.0.0"
+    "svelte": "^3.0.0",
+    "svelte-loader": "^2.13.6"
   },
```
robhudson, line 18: We should also run `npm update` on these. E.g. svelte is at 3.9.2. :exclamation:

robhudson:
This looks awesome. I'd say we should:
* Update the npm dependencies and remove unused dependencies as part of this merge.
* Remove the bundle.* files and add them to the `.gitignore`.

When those are done we should merge. :tada:

PR #13, opened 2019-09-06
Fix #11: Update to use probe API

### LINKED ISSUE #11: Update front-end to read probe data from API

robhudson:


src/app/sections/TelemetrySearchResults.svelte
```diff
@@ -115,21 +115,20 @@ li:hover {
     </div>
     {#if $searchResults.results.length}
         <ul>
-        {#each $searchResults.results as { name, probeType, description, id, searchID,
-        firstRelease, firstReleaseName }, i (id)}
+        {#each $searchResults.results as probe, i (probe.id)}
```
hamilton: code nit: could we destructure `probe` like `{#each $searchResults.results as {name, type, description, ...}, i (id) }`? I find it helps the readability in the block.

hamilton (changes requested):
Code looks good, with one minor nit. Two other questions that can be addressed separately:

(1) you had mentioned this over zoom, but I am encountering a CORS issue testing locally (that is, running from `:5000` from `npm run dev` and `:8000` from docker). We should probably include some docs about how to fix that or just have a `make dev` or something like that (I have usually just served the assets from whatever is responsible for the data as well). I'm guessing you probably tested it and it works great, but I haven't actually manually tested it yet.
(2) `telemetrySearch` could probably use some tests + some error-handling to surface into the frontend. For instance, the error I encountered in (1) was silently swallowed. One way I typically have handled this is to catch the network error in the readable and `set` it in a way that we can put it in the body of the dashboard. In this case, we can probably have `$telemetrySearch` always have a `status` key-value pair where the default is `{status: 'LOADING`. If `status === 'ERROR` we also have a `reason` k-v pair or something like this. This can definitely be a follow-on issue (we can, for instance, add Jest + write some quick tests + mocks around the callback into the `telemetrySearch` readable, not to mention build out required components) but we'll want to make sure that when there's an error people either know _what_ is happening or _who to bug_ (which will be us).

src/app/sections/TelemetrySearchResults.svelte
robhudson: There's 9 properties by my count which would make that a long line. And we're not using all properties in the block. We would also have to "restructure" all the properties back again in the `updateProbe` call (or only those we want in the store, but having what's in the API in the store is good, I think). I find having the object and the dot notation to access the properties very readable here.

src/app/store/telemetry-search.js
```diff
@@ -1,37 +1,22 @@
 import { readable } from 'svelte/store';
 
 import FlexSearch from 'flexsearch'
-const url = 'https://probeinfo.telemetry.mozilla.org/firefox/all/main/all_probes'
+// TODO: Make this dynamic based on prod vs local dev.
```
robhudson, line 6: What are you thoughts on achieving this? Is there a pattern for this?

src/app/store/telemetry-search.js
```diff
@@ -1,37 +1,22 @@
...
-        name: info.name, probeType: info.type, description, id, firstRelease, firstReleaseName
-    }
-    return out;
-}
 
 const telemetrySearch = readable({loaded: false}, async (set) => {
-    let data = await fetch(url).then(r=>r.json())
-    data = Object.values(data).map(resultToProps)
+    const resp = await fetch(url).then(r=>r.json())
+    // TODO: Change API to return array of probe objects?
```
robhudson: @hamilton What's your vote? Would array of objects vs one big object be more API user friendly?

robhudson:
> I am encountering a CORS issue testing locally

Make sure you run `make build` again so it picks up the new Python lib dependency and adds a new copy of the code. That fixed the CORS issue for me.

Great ideas on the error surfacing. It could be nice if there were a UI element we can update when things are loading from the network or errors happen, to keep the users informed either way.

src/app/sections/TelemetrySearchResults.svelte
hamilton: I think we're only using a handful specifically in the results (usually I only destructure the things I need), but I also recognize this is a matter of taste / eslint-in-react-land preference.

src/app/store/telemetry-search.js
hamilton: it seems you've actually taken care of this, right? `make build` should handle the cors stuff for this docker instsance

src/app/store/telemetry-search.js
hamilton: I think how we have it now is fine, and we can change down the line. This pattern of mapping the keys / spreading the values is pretty performant in practice.

src/app/store/telemetry-search.js
hamilton: ah, actually. I would just use `/api/v1/probes`. That should make the frontend indifferent to the domain.

robhudson: Using `/api/v1/probes` wouldn't handle the `:8000` (front-end) to `:5000) (back-end) host change. I imagine some set of constants defined for a locally running instance vs a production running instance.

src/app/sections/TelemetrySearchResults.svelte
robhudson: I'm not typically a front-end dev but destructuring is preferred here so I'll update it. I also just learned that you can destructure just the properties you want.

PR #19, opened 2019-09-11
Add eslint svelte plugin for `.svelte` files

robhudson:


hamilton:
@robhudson worked through it – looks good to me. I'd say let's merge it and if it causes anguish in some way we can tweak the rules

hamilton:
@robhudson on second thought, let's wait until end of M4 to do this in a final sweep.

PR #24, opened 2019-09-16
search component refinements

hamilton:
This draft PR implements the following:

- [x] keyboard interactions for search component & results
- [x] improved styling & semantics for search result list items
- [x] tweaks to the search function for speed and distance
- [x] some kind of basic spinner while search results are being fetched
- [-] ~refinements to svelte components / generalization~

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,8 +10,7 @@
     "extends": [
         "airbnb-base"
     ],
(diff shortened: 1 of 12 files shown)


hamilton:
@robhudson I think I'm going to hold off generalizing this component for reusability until after we're done with this sprint.

PR #28, opened 2019-09-18
Add Python import via dataproc and pyspark

robhudson:


dataproc/dataproc.py
```diff
@@ -0,0 +1,167 @@
...
+    job = bigquery_client.query(query)
+    job.result()  # Wait for job to complete.
+
+    logger.info("Query complete, processing via Spark.")
+
+    (
+        SparkSession.builder.appName("export_to_firestore")
+        .getOrCreate()
+        .read.format("bigquery")
+        .option("parallelism", 6)
```
relud: this should probably be set to 0, which lets the api decide based on volume, instead of a static number. if that doesn't work, try it at a variety of numbers, like 20, 50, and 100, and pick the highest number that reliably completes. I would expect at least 20 to work reliably. I found that setting it too high resulted in Out of Memory exceptions for main summary, but that job had complications that this might not.

relud: alternately, make this config option, so that it doesn't block merging the PR.

relud (approved):
r+wc

Issue #8, opened 2019-09-06
Create a single file that defines filter fields


hamilton:
In GUD, we have a single json file that determines the different values for filters & selectors. This drastically simplifies the process of adding new fields + filters without requiring knowledge of the underlying code.

PR #30, opened 2019-09-19
WIP: creates config.json, connects probe selection to actual data pull

### LINKED ISSUE #25: connect search selection to API data fetch

hamilton:
This will close #25 & #8. This PR will:

- [x] create a new `config.json` file, which defines fields, descriptions, etc. for queries
- [x] adds a client-side API endpoint, `api.js`, which is in charge of querying firestore for probe data
- [x] connects the new `config.json` file directly to the store
- [x] adds a new derived store which takes as input the app store and outputs a data set asyncronously

I will add more to this list that has not been completed.


api/app.py
```diff
@@ -88,7 +88,7 @@ def get_data():
         raise APIException("Missing JSON body or proper mimetype")
 
     q = body["query"]
-
+    print(q)
```
hamilton: strange – this was removed in a recent commit but does not appear to be rflected here

src/app/sections/GraphicBody.svelte
```diff
@@ -1,18 +1,24 @@
...
 setContext('updateSearchIsActive', store.connect(updateSearchIsActive));
+
 </script>
 
 <style>
 
 .graphic-body-container {
     padding: var(--space-2x);
+    overflow-y: auto;
+    height: calc(100vh - var(--header-height) * 2 - var(--space-4x));
```
hamilton, line 18: this is to enforce a more app-like overflow in the body when necessary

src/app/store/api.js
```diff
@@ -0,0 +1,14 @@
+const URL = 'http://localhost:8000/api/v1/data';
+
+/* eslint-disable import/prefer-default-export */
+export async function getProbe(params) {
```
robhudson: This is called `getProbe`, but the `URL` is for the histogram data, not just the probe. Did you want the probe URL -- `/api/v1/probe/<name>`?

src/app/store/store.js
```diff
@@ -1,18 +1,50 @@
...
     audienceSize: 0,
     totalSize: 0,
   },
-  product: 'all',
-  channel: 'all',
-  os: 'all',
+  product: 'Firefox',
+  channel: getDefaultFieldValue('channel'),
+  os: getDefaultFieldValue('os'),
+  version: 69,
```
robhudson, line 46: Would this be a TODO to fetch the latest firefox versions for each channel? Is that info we should keep in firestore?

src/app/store/store.js
```diff
@@ -69,27 +92,67 @@ export const resetFilters = () => async () => {
...
+export const hasDefaultControlFields = derived(store, ($store) => Object.values(CONFIG.fields)
+  .every((field) => field.values[0].key === $store[field.key]));
+
+// ///// probe querying infrastructure.
+
+// hook into store here.
+
+function getParamsForDataAPI(obj) {
+  return {
+    version: `${obj.version}`,
```
robhudson: Why is this one not like the others?

src/app/store/api.js
hamilton: oh good point. This should be `getProbeData` probably.

src/app/store/store.js
hamilton: yep, will add a `TODO` here. I haven't figured out how we should handle the versioning quite yet.

src/app/store/store.js
hamilton: I hit some issues with it not being a string before, but it may have actually been something else causing the problem, not this.

PR #41, opened 2019-10-18
Add Dockerfile and serve static assets via flask

robotblake:
* Adds static asset serving to the flask app
* Creates a multi-stage Dockerfile for the different pieces
  * `frontend` builds the frontend assets
  * `backend` builds the python requirements in a virtualenv
  * `final` contains the assets, virtualenv, and the flask app


api/app.py
```diff
@@ -215,6 +217,11 @@ def get_probe(name):
         return APIException("Probe not found", 404)
 
 
+@app.route("/")
+def hello():
```
robhudson: We could name this `index`.

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM node:lts-slim AS frontend
+
+WORKDIR /app
+
+COPY package.json /app/
+RUN npm install
+COPY . /app/
+RUN npm run build
+
+FROM python:3.7-slim
```
robhudson: If we add the `AS` I think we can then have a docker compose file with a `target` for local development of the backend.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+WORKDIR /app
+
+COPY api/requirements.txt /app/
+RUN pip install -U pip \
+    && pip install -r requirements.txt
```
robhudson: This could be further optimized by building the Python libs in a builder image and copied to the final image.

I kind of like the basket project's approach of using `python -m venv venv/`: https://github.com/mozmeao/basket/blob/master/Dockerfile

robhudson:
Looks great! I'll base my refactoring work on these changes as well.

robhudson:
This PR was used as the basis for the Django branch, which has now been merged.

Closing

PR #53, opened 2019-10-24
add extra probe details

spasovski:
image

### DIFF
--- a/src/app/sections/ProbeDetails.svelte
+++ b/src/app/sections/ProbeDetails.svelte
@@ -48,16 +48,16 @@ h2 {
 
 .empty-details {
     height: 200px;
-    display:grid;
+    display: grid;
     place-items: center center;
-    font-weight:900;
+    font-weight: 900;
     font-size: 1.2em;
-    text-align:center;
+    text-align: center;
     padding: var(--space-2x);
     color: var(--cool-gray-400);
 }
 
-.probe__description {
+.probe-description {
     color: var(--subhead-gray-02);
 }
 
@@ -71,7 +71,7 @@ h2 {
 }
 
 .spinner-and-text {
-    text-align:center;
+    text-align: center;
     color: var(--cool-gray-400);
 }
 
@@ -80,16 +80,48 @@ h2 {
 }
 
 .probe-labels {
-    padding:0;
-    display:grid;
-    grid-auto-flow:column;
+    padding: 0;
+    display: grid;
+    grid-auto-flow: column;
     align-items: center;
     padding-left: var(--space-2x);
     padding-right: var(--space-2x);
     grid-column-gap: var(--space-2x);
-
     height: var(--increment);
+}
 
+.drawer-header {
+    background: transparent;
+    border: 0;
+    color: var(--body-gray-02);
+    height: auto;
+    margin-bottom: var(--space-base);
+}
+
+.details-list {
+    margin: 0;
+    padding: 0;
+    list-style: none;
+}
+
+.details-list > * {
+    margin: 0 0 var(--space-base) 0;
+}
+
+.detail--indented {
+    position: relative;
+    padding-left: var(--space-2x);
+    font-size: var(--text-015);
+}
+
+.detail--indented::before {
+    content: ">";
+    position: absolute;
+    height: var(--space-base);
+    width: var(--space-base);
+    top: 0;
+    left: 0;
+    color: var(--cool-gray-400);
 }
 
 </style>
@@ -97,8 +129,8 @@ h2 {
 <RightDrawer visible={paneVisible}>
 {#if !$telemetrySearch.loaded}
     {#if visible}
-    <div in:fly={rightDrawerTransition} class=drawer-section>
-        <div class=spinner-and-text>
+    <div in:fly={rightDrawerTransition} class="drawer-section">
+        <div class="spinner-and-text">
             <LineSegSpinner size={48} color={'var(--cool-gray-400)'} /> 
             <div in:fade={{ duration: rightDrawerTransition.duration * 2 }}>Loading Probes</div>
         </div>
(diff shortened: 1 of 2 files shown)


src/app/sections/ProbeDetails.svelte
```diff
@@ -122,26 +170,53 @@ h2 {
...
     </div>
     {/if}
-    <div class='drawer-section drawer-section--end'>
-        <h2 class=detail__heading--01>export</h2>
+    <div class="drawer-section">
+        <h2 class="detail__heading--01">extras</h2>
+        <dl class="details-list">
+        {#if $store.probe.active !== undefined}
+            <dt><span class="label label-text--01 label--definition-title">active</span></dt>
+            <dd class="details-list-definition label-text--01">{$store.probe.active}</dd>
```
hamilton: I like the use of `dd` and `dt` – let's do more of this in the future – but I'm a bit hesitant to go with this label color. At the moment, histograms have one color (this one) and scalars have another (red). We maintain that particular distinction in the search results as well. I'm not sure if this color should match that, or if we should just go with something a bit more neutral like `--blue-slate-xxx`.

src/app/sections/ProbeDetails.svelte
```diff
@@ -122,26 +170,53 @@ h2 {
...
+        {#if $store.probe.prerelease !== undefined}
+            <dt><span class="label label-text--01 label--definition-title">pre-release</span></dt>
+            <dd class="details-list-definition label-text--01">{$store.probe.prerelease}</dd>
+        {/if}
+        </dl>
+    </div>
+    {#if $store.versions && $store.versions.length}
+    <div class="drawer-section">
+        <header class="drawer-header drawer-header--split">
+            <h2 class="detail__heading--01">available in versions</h2>
```
hamilton: We might not need the word "versions" and at any rate I think it starts to crowd out the Beta that comes after.

src/app/sections/ProbeDetails.svelte
hamilton: Furthermore, I think we should probably think of breaking out label styling into two distinct things: normal (which is what we have here) and compact. Compact (which I think we should change this to) should probably have smaller left / right padding, and the text size should be smaller than the normal one. Otherwise it's hard to know what my eyes are supposed to make of this visual hierarchy. I'm open to removing the label entirely and just using overline styling or something like that here as well.

hamilton: More to this point, it'd be great if we could just say `inactive` here, and make the label bg light `--pantone-red-xxx` with darker text `--pantone-red-700` or something like that. "active: false" feels like too direct mapping from the data to the frontend, and I think people would just appreciate "inactive". Similarly, "active" could use `--mantis-green-xxx` with similar shade values. Same probably goes with prerelease – maybe we just say "unavailable in prerelease" or similar copy

src/app/sections/ProbeDetails.svelte
```diff
@@ -122,26 +170,53 @@ h2 {
...
+            <a
+            href='https://bugzilla.mozilla.org/show_bug.cgi?id={bugID}'
+            target="_blank">{bugID}</a>
+        {/each}
+        </div>
     </div>
     {/if}
-    <div class='drawer-section drawer-section--end'>
-        <h2 class=detail__heading--01>export</h2>
+    <div class="drawer-section">
```
hamilton: I would consider combining "extras" with the "available in" sections, with a goal to make the data labeling in the sections speak enough for itself.

src/app/sections/ProbeDetails.svelte
```diff
@@ -122,26 +170,53 @@ h2 {
...
+        </dl>
+    </div>
+    {#if $store.versions && $store.versions.length}
+    <div class="drawer-section">
+        <header class="drawer-header drawer-header--split">
+            <h2 class="detail__heading--01">available in versions</h2>
+            <span class="label-text--01">{$store.channel}</span>
+        </header>
+        <dl class="details-list">
+            <dt><span class="label label-text--01 label--definition-title">first</span></dt>
```
hamilton: similarly, we might just say `beta     47 - 70` here instead of having the first, last, and beta info as separate entities

hamilton (changes requested):
Looking good @spasovski! I love the choices you made w/ the html & the cleaning up of the css is much-needed. I left a few comments about how we might think about actually organizing the information. We have a great opportunity here to go beyond just mirroring the individual bits of data directly. Let me know what you think.

src/app/sections/ProbeDetails.svelte
spasovski: Really good points. I wasn't sure of the color plan moving forward so I just picked something random but more importantly I really agree with just "active" vs "inactive" rather than writing out the booleans.

spasovski:
Great suggestions - I still like this approach of making some choices and then discuss if any obvious improvements come to surface.

src/app/sections/ProbeDetails.svelte
```diff
@@ -80,25 +80,73 @@ h2 {
...
+}
+
+.drawer-header .detail__heading--01 {
+    padding: 0;
+}
+
+.drawer-header--split {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
```
spasovski: good candidate for a mixin later

src/app/sections/ProbeDetails.svelte
hamilton: no worries re: the color – I haven't communicated much of that! I'm thinking this would be probably a good use of Storybook to demonstrate / agree on the colors for GLAM, actually. That way we can come to a consensus about these details that currently are unlocked in my head, unchallenged.

spasovski:
@hamilton just pushed some changes, removed the labels for now...it might have been too much visual emphasis on minor meta details but open to discussion

hamilton (approved):

hamilton:
Looks good – thanks @spasovski! Merging.

PR #110, opened 2019-11-20
don't use localhost:8000 on production builds

hamilton:


package.json
```diff
@@ -50,7 +51,7 @@
   },
   "scripts": {
     "build": "rollup -c",
-    "autobuild": "rollup -c -w",
+    "autobuild": "rollup -c -w NODE_ENV:production",
```
robhudson: I think this requires the `--environment` flag

src/app/state/api.js
```diff
@@ -1,8 +1,10 @@
 /* eslint-disable import/prefer-default-export */
 
+const path = '__BASE_DOMAIN__';
+const url = `${path}/api/v1/data/`;
```
robhudson: You could also do the following, right, and skip the extra variable?
```
const url = '__BASE_DOMAIN__/api/v1/data/';
```

rollup.config.js
```diff
@@ -17,6 +18,7 @@ export default {
   },
   plugins: [
     json(),
+    replace({ __BASE_DOMAIN__: production ? '' : 'http://localhost:8000' }),
```
robhudson, line 21: This uses the `ROLLUP_WATCH` from line 9 instead of the one you added in the build script. I think to use that you'd instead do:
```
     replace({ __BASE_DOMAIN__: process.env.NODE_ENV === 'production' ? '' : 'http://localhost:8000' }),
```
??

package.json
hamilton: oops, need to remove it, not actually necessary

src/app/state/api.js
hamilton: I suppose so :)

robhudson (approved):

PR #126, opened 2019-11-21
(fixes #115) conditional render of sidebar

### LINKED ISSUE #115: hide/remove the sidebar on the "home" view
The home page view has no need of the sidebar - it should be removed and re-added on the probe details view. This may need to be done via "visibility: hidden" so that it still takes up space.

spasovski:
image


hamilton (changes requested):
This looks good! The right side drawer is still using a transition that I wrote long ago when I thought we were going full-width. Could you remove that from the right drawer component? It feels a bit abrupt to me. Let me know if you need help with tackling that.

spasovski:
@hamilton updated - seems to work fine locally

hamilton (approved):
Yep, this works great now. Thanks @spasovski

PR #130, opened 2019-11-21
(fixes #117) column layout for probe type

### LINKED ISSUE #117: probe type and subtype should be restyled
The decision reached was to have the main type and subtype be arranged each in its own row (like `display:flex;flex-direction: column;`) with the main type in a bigger font-size. They can still live in the current visual space.

spasovski:
image

### DIFF
--- a/src/app/sections/ProbeDetails.svelte
+++ b/src/app/sections/ProbeDetails.svelte
@@ -9,6 +9,7 @@ import { store, dataset } from '../state/store';
 import { downloadString } from '../../utils/download';
 
 import Button from '../../components/Button.svelte';
+import StatusLabel from '../../components/StatusLabel.svelte';
 
 const paneTransition = { x: 10, duration: 300 };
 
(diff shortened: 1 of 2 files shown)


src/app/sections/ProbeDetails.svelte
```diff
@@ -149,18 +158,12 @@ h2 {
...
     <div class="probe-details-content">
         {#if $store.probe.type}
-            <div class="drawer-section probe-labels">
-                <div>
-                    <span
-                        style="display: inline-block;"
-                        class="label label-text--01 label--{$store.probe.type}">{$store.probe.type}</span>
-                </div>
+            <dl class="drawer-section info-type">
+                <dt>{$store.probe.type}</dt>
```
hamilton, line 153: the `dt` here looks a little similar to the headers in the other parts of the details section. Could we play with making it not using uppercase + making it a bit bigger?

hamilton (changes requested):

spasovski:
@hamilton updated - some of the naming may not be ideal but open to suggestions

src/app/sections/ProbeDetails.svelte
```diff
@@ -99,34 +89,53 @@ h2 {
...
-.detail--indented::before {
-    content: ">";
-    position: absolute;
-    height: var(--space-base);
-    width: var(--space-base);
-    top: 0;
-    left: 0;
-    color: var(--cool-gray-400);
+.info-type--padded {
+    padding: 0 9px;
```
hamilton:
```suggestion
    padding: 0 var(--space-base);
```

hamilton: I believe `var(--space-base)` is 8px, does this work ok?

src/app/sections/ProbeDetails.svelte
```diff
@@ -99,34 +89,53 @@ h2 {
...
+}
+
+.info-type--subtle {
+    display: flex;
+    justify-content: space-between;
+    padding-top: 0;
```
hamilton, line 112:
```suggestion
    padding-top: 0;
    align-items: baseline;
```

hamilton: the `align-items: baseline` will make sure the vertical alignment matches b/t the number and the channel. By default I think it is `start` which looks off

src/components/StatusLabel.svelte
```diff
@@ -0,0 +1,42 @@
...
+
+<style>
+
+.status-label {
+  font-size: var(--text-015);
+}
+
+.status-label span {
+  display: inline-block;
+  padding: 3px 9px 4px;
```
hamilton:
```suggestion
  padding: var(--space-1h) var(--space-base) var(--space-1h);
```

hamilton: this might be controversial, but keeping these values closedly aligned to the spacing vars will help us in the long-term. Let me know if it looks weird like this. I would also be open to `calc(var(--space-base) + 1px)` if it feels necessary to have these pixels closer to what you have. I know it's kind of verbose but it feels right to stick to the css vars we're using, but maybe this is too strict. Open to counterargument here

src/components/StatusLabel.svelte
```diff
@@ -0,0 +1,42 @@
...
+  border-radius: 11.5px;
```
hamilton: This border radius feels like an argument for something like `calc(var(--space-base) * 1.5)` which is `12px` in the current formulation. Is there a reason for `11.5px` vs. `12px`?

src/app/sections/ProbeDetails.svelte
```diff
@@ -99,34 +89,53 @@ h2 {
...
+.info-type--subtle dt {
+    text-transform: none;
+    font-size: var(--text-02);
+}
+
+.info-type--subtle dd {
+    font-size: var(--text-01);
+}
+
+.probe-details-intro {
```
hamilton:
```suggestion
.probe-details-overview {
```

src/app/sections/ProbeDetails.svelte
```diff
@@ -145,39 +154,32 @@ h2 {
 <div in:fly={paneTransition} class="drawer-section-container probe-details">
     <!-- probe-details-content -->
     <div class="probe-details-content">
+        <div class="probe-details-intro">
```
hamilton:
```suggestion
        <div class="probe-details-overview">
```

src/app/sections/ProbeDetails.svelte
```diff
@@ -145,39 +154,32 @@ h2 {
...
                 {#if $store.probe.kind}
-                    <div class="probe-kind label-text--01">
-                        {$store.probe.kind}
-                    </div>
+                    <dd>{$store.probe.kind}</dd>
                 {/if}
+            </dl>
+        {/if}
+        {#if $store.probe.active !== undefined}
+            <div class="info-content">
```
hamilton: I'm wondering if we might be able to find more semantic names for `info-type` and `info-content` since semantically these bits of information are quite similar. Seems like they're just sort of left / right overview content, since it's less about whether it's `type` or `content` and more about handling the positioning. What do you think?

hamilton (changes requested):
This is getting close @spasovski. Left a few comments & suggestions – let me know what you think

src/components/StatusLabel.svelte
spasovski: Yep agreed - visually the values that work best are not symmetric but they *should* have a variable base we can offset from.

spasovski:
@hamilton ready for re-r when you get a chance

hamilton (approved):
looks good – thanks!

PR #133, opened 2019-11-26
(fixes #158)(fixes #120)(fixes #122) restyle content columns

### LINKED ISSUE #120: de-emphasize the sidebar
This has no concrete definition and will likely result in some back-and-forth discussion. A good first try would be to lessen the background-color opacity like `background: rgba(255,255,255, 0.3)` and border lines.

### LINKED ISSUE #122: consider more visual emphasis on content area
This is quite loose and will likely be less styling work than it is discussion. Consider adding a border, border-radius, box-shadow etc.

### LINKED ISSUE #158: change description column's detail-heading--01 to bolder
Bump from 100 to 200

spasovski:
It's probably better to pull this locally for preview - I didn't want to screenshot too much data even if it isn't real. This is obviously one attempt - we may want to iterate a few times until we agree on a strategy.

image

### DIFF
--- a/src/app/sections/GraphicBody.svelte
+++ b/src/app/sections/GraphicBody.svelte
@@ -93,36 +93,33 @@ function handleBodySelectors(event) {
 <style>
 
 .graphic-body-container {
-    display:grid;
-    grid-template-columns: auto min-content;
-    grid-template-rows: auto auto;
-    grid-template-areas: "header header"
-                         "content-body right";
-    align-items: stretch;
-    justify-items: stretch;
-    background-color: white;
-    --height: calc(100vh - var(--header-height) * 2);
+  display: grid;
+  grid-template-columns: auto min-content;
+  grid-template-rows: auto auto;
+  grid-template-areas: "header header";
+  align-items: stretch;
+  justify-items: stretch;
+  --height: calc(100vh - var(--header-height) * 2);
 }
 
 .graphic-body__graphic-header {
-    grid-area: header;
-    display: grid;
-    grid-template-columns: auto max-content;
-    grid-column-gap: var(--space-4x);
-    align-items: start;
-    background-color: var(--cool-gray-100);
+  grid-area: header;
+  display: grid;
+  grid-template-columns: auto max-content;
+  grid-column-gap: var(--space-4x);
+  align-items: start;
+  background-color: var(--cool-gray-100);
 }
 
 .graphic-body__graphic-header h2 {
-    margin:0;
-    padding:0;
-    width: 100%;
-    word-break: break-all;
-    height: var(--increment);
-    display: grid;
-    align-items: center;
-    padding-left: var(--space-4x);
-    
+  margin: 0;
+  padding: 0;
+  width: 100%;
+  word-break: break-all;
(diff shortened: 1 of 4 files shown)


src/app/sections/GraphicBody.svelte
```diff
@@ -126,6 +125,8 @@ function handleBodySelectors(event) {
   background-color: white;
   padding: var(--space-4x);
   padding-top: var(--space-2x);
+  border-radius: var(--border-radius-1h);
+  box-shadow: 0 0 3px 2px rgba(0,0,0,0.1) inset;
```
hamilton:
```suggestion
  border: 1px solid var(--cool-gray-100);
```

hamilton: The inset shadow is kind of intense and might be a bit distracting. I'd maybe recommend:

1. removing that,
2. adding a 1px solid border to the body element

hamilton (changes requested):
Thanks for the PR! I agree that we'll probably have to have some back-and-forth on this. Here are some thoughts on where it is now:

1. I love the bg color to the sidebar – definitely helps separate that bit of content adn de-prioritizes it.
2. the inset feels a bit intense, I think. We might be able to accomplish a bit of separation by having a 1px solid right border on the  body.
3. it seems like it might help if we have the details + body in its own element + apply a light box shadow to that instead of the grid approach we have here. That'll help with keeping the body content as a whole the most important element on the screen, and still maintain the de-prioritized details.

Let me know what you think ~

spasovski:
@hamilton re-r? Likely needs another few iterations of comments but no big deal.

hamilton (changes requested):
Thanks for the tweaks. A few comments that we can talk about more tomorrow:

- the depth variable on the shadow seems to only apply to `graphic-body__content`. I think the solution here is to have a parent element that encompasses `graphic-body__content` and `graphic-body__details` (as well as another grid for that new parent element) and you can apply the shadow variable to that. As it is now the difference between `--subtle-cool` and the bg color is pretty minimal, and we aren't achieving any depth on the right side there. I'm open to changing the feel of the depth variable if it feels "too deep", but we should probably get the styling there consistent either way.
- It does still feel that maybe a 2px solid border b/t the body & the right column might help separate things a bit. I found `--cool-gray-100` to look pretty good. Worth playing with.
- it definitely feels like the top two lines (histogram nightly active ###) could be a different color, or maybe a slight linear gradient like `linear-gradient(to top right, white 50%, var(--cool-gray-100))`. Not sure, but as it is now the label element you created doesn't feel like it works well against the background.

spasovski:
@hamilton should be ready for next round of review

- adjusted a11y contrast of StatusLabel states
- added letter-spacing + font-weight: 200 on detail-header__01

hamilton:
@spasovski thanks for the changes. This is looking great. I am noticing that the spacing you added for the header creates a scrollbar where there wasn't one before. Let's either fix that or reverse the change and address it later. Should be a quick fix either way – once it's done, feel free to merge this.

spasovski:
@hamilton awesome thanks and yea I'll just undo it and have it go in its own PR

PR #137, opened 2019-11-27
Implement auth0 as DRF authentication class

### LINKED ISSUE #91: add basic authentication to the frontend
The frontend will need to establish authentication. For. v0.1a we are hoping to just making it require SSO by default, with v0.2a allowing some version of unauthenticated access for some subset of API calls / params.

For now, this should entail:

- add something to the store that contains auth information
- have the API requests to the server be signed w/ the appropriate auth information from the store
- ~write the appropriate functionality to handle the auth UX as a Svelte component (a big "log in" button will suffice for this issue, nothing fancy)~  (this might not be necessary if you have to authenticate to even see the dashboard)

robhudson:
closes #91


openjck:
Added front-end authentication support. The front-end will not load unless the user authenticates. I don't believe we can test authenticating with the API, however, until we are provided [a value for FRONT_END_AUTH_AUDIENCE](https://bugzilla.mozilla.org/show_bug.cgi?id=1605517).

None of the new environment variables are secrets. They should be set to:

```
FRONT_END_AUTH_DOMAIN=auth.mozilla.auth0.com
FRONT_END_AUTH_CLIENT_ID=Doy3pSTMj7yfSP2yaVpIcCNPcXtubdDh
FRONT_END_AUTH_AUDIENCE=[TBD]
```

I'm not exactly sure how to test the protected parts of the API, since this is my first PR to GLAM. If the `gc_ms` data loads when the *Release* channel is set, would it be fair to say that this is working as expected? What other actions can be taken to test that this works correctly?

openjck:
@robhudson You may want to double-check that I did the merge correctly. The only Python-related file which conflicted was *requirements.txt* and I think I've merged it correctly.

Apologies for the force push a moment ago. I wanted to remove an earlier merge which contained a mistake, but I know this is a collaborative branch so I won't force push again in the future.

src/app/state/api.js
```diff
@@ -2,11 +2,12 @@
 
 const url = '__BASE_DOMAIN__/api/v1/data/';
 
-export async function getProbeData(params) {
+export async function getProbeData(params, token) {
   const data = await fetch(url, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
+      Authorization: `Token ${token}`,
```
robhudson: This should be a `Bearer` token. The backend is looking for that string.

robhudson:
> I'm not exactly sure how to test the protected parts of the API, since this is my first PR to GLAM. If the gc_ms data loads when the Release channel is set, would it be fair to say that this is working as expected?

Yep, anything with `?channel=release` will check for authentication.

glam/api/views.py
```diff
@@ -84,9 +80,14 @@ def aggregations(request):
             "Missing required query parameters: {}".format(", ".join(sorted(missing)))
         )
 
+    # If release channel, make sure the user is authenticated.
```
hamilton, line 27: this will have to get rebased on master since this view has changed a bit, but I don't think it'll be too difficult

src/utils/auth.js
```diff
@@ -0,0 +1,47 @@
...
+
+    const isAuthenticated = await auth0.isAuthenticated();
+
+    if (isAuthenticated) {
+      withToken(successCallback);
+    } else {
+      const query = window.location.search;
+      if (query.includes('code=') && query.includes('state=')) {
+        await auth0.handleRedirectCallback();
+        window.history.replaceState({}, document.title, '/');
```
hamilton: help me understand – will this remove the query string if it existed before? At the moment, GLAM & GUD work similarly in that we are capturing the relevant parts of the app state in the url. It looks like the redirect done this way might clobber that.

src/main.js
```diff
@@ -1,4 +1,10 @@
 import TelemetryApp from './app/TelemetryApp.svelte';
+import { authenticate } from './utils/auth';
+import { store } from './app/state/store';
```
hamilton: wondering if it might make sense to (1) have a `createStore` function a la GUD here, and then (2) have the store instantiation happen in the `authentication` success callback. I suspect we'll have patterns like this in other data products down the line so it'd be good to get this right now rather than later. As it stands now, GLAM's state management is pretty close to GUD's as it is, but the idea of making a `createStore` function came out of our refactor, and hasn't hit this project yet.

src/app/TelemetryApp.svelte
```diff
@@ -24,29 +24,27 @@ function updateQueryString() {
...
-                <TelemetryAppBar />
-                <Search />
-                <TelemetryMainFilters />
-        </MainActionBar>
-        <Content>
-          <GraphicBody />
-        </Content>
-    </Main>
-</App>
+{#if $store.token}
```
hamilton: is the idea here that if `$store.token` doesn't exist, we won't see anything at all?

src/app/TelemetryApp.svelte
```diff
@@ -24,29 +24,27 @@ function updateQueryString() {
   }
 }
 
-$: if (visible) {
+$: if ($store.token && visible) {
```
hamilton: is it necessary to check `$store.token` here?

hamilton (commented):
Did a preliminary look over the code & left a few comments. Let me know if anything is unclear

src/app/state/api.js
openjck: Thanks!

src/utils/auth.js
openjck: Ah, yes. I took this from the tutorial thinking that it'd be fine (it'll only happen when the user first visits the site), but this will also break links and bookmarks. Good call. Thank you!

src/app/TelemetryApp.svelte
openjck: Yep, exactly. I'll add a comment. :smiley:

src/app/TelemetryApp.svelte
openjck: I think so. If `$store.token` is removed, `updateQueryString` does run when the user is unauthenticated. It doesn't seem to cause any harm, but it does seem unnecessary. Open to suggestions. :smiley:

src/main.js
openjck: GUD doesn't have a `createStore` function. Do you mean `getDefaultState`?

openjck: Ah. GLAM does have a `createStore` function already. I think I see what you mean. I'll move it there.

openjck: I think I've done what you suggested, but let's check back in on this.

openjck: I'm going to revert my work on this. I tried the following:

### store.js

```javascript
[...]
export let store = {};
export function initStore() {
  store = createStore(initialState);
}
[...]
```

### main.js

```javascript
import TelemetryApp from './app/TelemetryApp.svelte';
import { authenticate } from './utils/auth';
import { store, initStore } from './app/state/store';

// eslint-disable-next-line import/no-mutable-exports
let telemetryApp;

authenticate((token) => {
  initStore();
  store.setField('token', token);

  telemetryApp = new TelemetryApp({
    target: document.body,
  });
});

export default telemetryApp;
```

But I was getting an error about `store.subscribe` not being a function. Is this along the lines of what you had in mind, at least?

hamilton: yep, something like this is what I had in mind, but I can see why this might not work. It might end up just making more sense to not do the store instantiation in the `authenticate` success callback but just do the `setField` there.

src/app/TelemetryApp.svelte
hamilton: will `$store.token` ever be removed?

openjck:
Rob Hudson and I are working on an issue related to the communication between the client and the server. Justin Dow shared some information today which may help to unblock us and we also have a meeting scheduled with him on Monday to address other questions we have.

I spoke with Hamilton and suggested that he not review this much more until this gets sorted out. I'll address any current feedback and do a merge now, though.

src/main.js
openjck: Can you say more about why that approach may not work? Is there something inherently wrong with it?

I'll leave this file as-is for now but I'm happy to revisit if we find a better way to do this.

src/app/TelemetryApp.svelte
openjck: Shouldn't be. It's set if the user authenticates and unset otherwise.

openjck:
Addressed feedback and merged *mozilla/master* in.

openjck:
Looks like a couple of tests are failing, but otherwise this should be ready for review.

hamilton:
hmm, pulled this branch to do a review tonight, but having trouble getting it to work properly. I have the secrets as per instructions. But going to `localhost:5000` redirects me to `https://null/authorize?client_id=null&audience=null&scope=openid profile email[and so on]`

src/main.js
hamilton: Yep, let's just leave this as-is for now. I think your approach works great here.

hamilton (changes requested):
This is looking great @openjck. Some feedback / questions for you:

(1) Some code has changed a bit from when you filed this PR that will probably need to be considered. You can see where this breaks if you click on the GLAM logo on the top left – it actually resets the store and sends you back to the main splash page, and it removes the token. In `store.js`, there is a funcfion called `store.reset` which kind of "resets" the store state to what it was on load. This is _not_ strictly what it does – there are a few things (like the probe value) that needs to be reset entirely, and some other values reset to a different default. But right now now, when `store.reset` gets called, the token gets destroyed. This can be easily fixed by getting the token from the store in that function and re-setting it after the reinitialization happens. Let me know if what I've described is clear. I think this should be a trivial fix.
(2) I notice that sometimes that when I reload, it takes sometimes 5-10 seconds. Is that expected behavior? Kind of surprising. Will this be necessary for local development? Should we consider at some point adding a loading spinner or something? I'm worried that waiting 5-10sec is kind of a non-starter for users.
(3) If someone goes to `<GLAM URL>?params` and they are not logged in, they will probably go through the auth workflow. Will `?params` be clobbered when they're redirected back to the URL? I'm not sure how to test that edge case. Any ideas there? We still want to parse the URL and get the bits of state out of it when the store initializes.

openjck:
1. Great catch. Fixed. Thanks!
2. You mentioned on Slack today that this may have been due to a a drop in your internet connection. I haven't noticed this happening to me, but I'll keep an eye out.
3. Another good catch, thanks! This has been fixed. It can be tested by loading a particular probe, visiting sso.mozilla.com, logging out, and then reloading Glam.

openjck:
I'm going to spend some time now trying to get the back-end tests fixed. If it takes too long, we can ask Rob how he feels about merging with broken tests. (Probably not great?)

openjck:
I believe the tests are failing because we don't have any data on CI. I'll ask Rob tomorrow.

openjck:
Rebased, although I'm seeing an issue on master and this branch about `App` not being defined. I also reached out to infosec to ask for their help in reviewing this.

PR #143, opened 2019-12-04
remove content area scroll, default to full page scroll

spasovski:
Was thinking this might break a bunch of things but so far everything seems fine locally.

### DIFF
--- a/public/static/global.css
+++ b/public/static/global.css
@@ -405,15 +405,18 @@ main .main-content {
     justify-self: stretch;
     display: grid;
     grid-template-columns: auto max-content;
+    padding-top: var(--space-6x);
 }
 
 main .main-action {
     background-color: var(--blue-slate-700);
-    position: relative;
+    position: fixed;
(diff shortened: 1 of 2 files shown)


hamilton (changes requested):
Thanks for taking this on. I think there's probably a bit more to be done here. A good test probe that will cause a scroll is `browser_engagement_navigation_urlbar`. See gif:

{gif removed}

1. we'll need to make sure to apply the body bg styling (and probably make it so there isn't some height on whatever that element is now)
2. We should probably make the app bar (the thing w/ the search) `position: sticky; top: 0` so that it remains at the top
3. there is a possibility that the right column should also have some kind of `position:sticky` as well, but I'm not sure. It might be nice for the detail content to remain visible when scrolling

hamilton (approved):
Thanks @spasovski – this looks good to me. I might make a follow-on here using `<svelte:window />` bindings but I think this sets me up to finish the table views.

PR #144, opened 2019-12-04
add header link to project root

spasovski:
🤷‍♂


hamilton (changes requested):
Thanks for the PR, and good catch. Could we remove the underline here on hover?

Sure would be nice to both have this as a link as you have here, but prevent the actual page refresh. I guess that'd require some router or something. The firebase console is using some routing when you click the logo that doesn't cause a whole-page reload. Thankfully our bundle is small anyway. Maybe we approach that after we feel good about what we have

Kapture 2019-12-04 at 19 01 42

PR #175, opened 2019-12-31
implement new cool-gray palette

spasovski:
I am sure it will need some more adjustments but it still feels quite close to the original.

### DIFF
--- a/public/static/global.css
+++ b/public/static/global.css
@@ -127,24 +127,34 @@
     --blue-slate-900: hsl(240, 25%, 9%);
 
     /* UX GRAYS – use the cool gray variables. */
-    --cool-gray-050: #f5f9fb;
-    --cool-gray-100: hsl(203, 55%, 96%);
-    --cool-gray-200: hsl(203, 40%, 89%);
-    --cool-gray-300: hsl(203, 30%, 82%);
-    --cool-gray-400: hsl(203, 20%, 71%);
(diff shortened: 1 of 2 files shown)


public/static/global.css
```diff
@@ -155,7 +163,7 @@
     --digital-blue-300: hsl(209, 85%, 75%);
     --digital-blue-400: hsl(209, 95%, 60%);
     --digital-blue-500: hsl(209, 100%, 39%);
-    --digital-blue-600: hsl(209, 100%, 34%);
+    --digital-blue-600: hsl(209, 77%, 46%);
```
hamilton: was this change intentional?

hamilton (changes requested):
Thanks for getting this filed. This is a good start, but I think there are some differences here that feel pretty stark:

- the background color is noticeably darker in the app.
- I'm wondering if the color set is a bit too neutral. If you compare what's at the internal staging site to what's in this branch (specifically the search functionality), you can see the diff (see the screenvid). The linear version might be a better bet.
- we should probably jump on Zoom to talk about this more and screen share and figure out what we'll need to tweak in the usage of the colors.

Let me know how you want to proceed here. I think maybe the sine curve for saturation actually makes things not cooled enough.

Kapture 2019-12-31 at 8 58 33

public/static/global.css
spasovski: nope, good catch

PR #179, opened 2020-01-03
creates new get_aggregations function and applies it to the aggregations view

### LINKED ISSUE #178: separate out the aggregation-fetching from the view
Currently, the `aggregations` view contains the logic for fetching the aggregation. Refactoring this code and putting it into a function will make it much easier to fetch aggregations in other views (for instance, the random probe view).

hamilton:
closes #178

### DIFF
--- a/glam/api/views.py
+++ b/glam/api/views.py
@@ -16,7 +16,7 @@ from glam.api.models import Aggregation, Probe
 
 
 def get_aggregations(**kwargs):
-    REQUIRED_QUERY_PARAMETERS = ["channel", "probe", "versions", "aggregation_level"]
+    REQUIRED_QUERY_PARAMETERS = ["channel", "probe", "versions", "aggregationLevel"]
 
     labels_cache = caches["probe-labels"]
     if labels_cache.get("__labels__") is None:
@@ -33,9 +33,9 @@ def get_aggregations(**kwargs):
         Q(metric=kwargs["probe"]),
         Q(channel=CHANNEL_IDS[kwargs["channel"]]),
         Q(version__in=map(str, kwargs["versions"])),
-        Q(os=kwargs["os"]),
+        Q(os=kwargs["os"] if "os" in kwargs else None),
     ]
-    aggregation_level = kwargs["aggregation_level"]
+    aggregation_level = kwargs["aggregationLevel"]
 
     # Whether to pull aggregations by version or build_id.
     if aggregation_level == "version":
@@ -154,28 +154,17 @@ def aggregations(request):
         }
 
     """
-    REQUIRED_QUERY_PARAMETERS = ["channel", "probe", "versions", "aggregationLevel"]
-
     body = request.data
 
     if body is None or body.get("query") is None:
         raise ValidationError("Unexpected JSON body")
 
-    q = body["query"]
-
-    if any([k not in q.keys() for k in REQUIRED_QUERY_PARAMETERS]):
-        # Figure out which query parameter is missing.
-        missing = set(REQUIRED_QUERY_PARAMETERS) - set(q.keys())
-        raise ValidationError(
-            "Missing required query parameters: {}".format(", ".join(sorted(missing)))
-        )
-
     response = get_aggregations(**body["query"])
 
-    # Strip out the merge keys when returning the response.
     if not response:
         raise NotFound("No documents found for the given parameters")
 
+    # Strip out the merge keys when returning the response.
     return Response(
         {
             "response": [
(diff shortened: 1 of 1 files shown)


glam/api/views.py
```diff
@@ -155,11 +101,89 @@ def aggregations(request):
...
+        raise ValidationError("Unexpected JSON body")
+
+    q = body["query"]
+
+    if any([k not in q.keys() for k in REQUIRED_QUERY_PARAMETERS]):
+        # Figure out which query parameter is missing.
+        missing = set(REQUIRED_QUERY_PARAMETERS) - set(q.keys())
+        raise ValidationError(
+            "Missing required query parameters: {}".format(", ".join(sorted(missing)))
+        )
```
robhudson: This is now both in this API view and in the Python function. We can probably remove it here and assume the `get_aggregations` handles these checks.

glam/api/views.py
```diff
@@ -155,11 +101,89 @@ def aggregations(request):
...
+                        "version": "70"
+                    }
+                }
+            ]
+        }
+
+    """
+    labels_cache = caches["probe-labels"]
+    if labels_cache.get("__labels__") is None:
+        Probe.populate_labels_cache()
```
robhudson: The `labels_cache` can be removed here since histogram labels now get attached in the `get_aggregations` method. No need to check this again.

glam/api/views.py
hamilton: oops, thanks for the check

glam/api/views.py
hamilton: yep, will remove, thanks for the catch

glam/api/views.py
```diff
@@ -155,11 +101,79 @@ def aggregations(request):
...
+    """
+
+    body = request.data
+
+    if body is None or body.get("query") is None:
+        raise ValidationError("Unexpected JSON body")
+
+    q = body["query"]
+
+    response = get_aggregations(
```
robhudson: You can probably just replace this with:

    response = get_aggregations(body["query"])

Since `get_aggregations` expects a dict, and then let it figure out if the required params are there.

glam/api/views.py
```diff
@@ -155,11 +101,79 @@ def aggregations(request):
...
+        aggregation_level=q["aggregationLevel"],
+        probe=q.get("probe"),
+        channel=q.get("channel"),
+        versions=q.get("versions"),
+        os=q.get("os"),
+    )
+
+    # Strip out the merge keys when returning the response.
```
robhudson: This comment doesn't match the code.

robhudson: It belongs down below with the `return Response`.

glam/api/views.py
```diff
@@ -15,86 +15,32 @@
...
-        Q(metric=q.get("probe")),
-        Q(channel=CHANNEL_IDS[q.get("channel")]),
-        Q(version__in=map(str, q.get("versions"))),
-        Q(os=q.get("os")),
+        Q(metric=kwargs["probe"]),
+        Q(channel=CHANNEL_IDS[kwargs["channel"]]),
+        Q(version__in=map(str, kwargs["versions"])),
+        Q(os=kwargs["os"]),
     ]
+    aggregation_level = kwargs["aggregation_level"]
```
robhudson: Tests fail b/c this should be `kwargs["aggregationLevel"]`

hamilton: this actually isn't why the test fails. It fails because `q["aggregationLevel"]` doesn't exist I think, not `kwargs["aggregation_level"]`. This worked before I pushed changes because I had the validation in the view and the `get_aggregations` function

glam/api/views.py
hamilton: 👍

glam/api/views.py
hamilton: I'm going to revert back to camel case in `get_aggregations` because this is causing noise

glam/api/views.py
```diff
@@ -15,86 +15,32 @@
...
 
     dimensions = [
-        Q(metric=q.get("probe")),
-        Q(channel=CHANNEL_IDS[q.get("channel")]),
-        Q(version__in=map(str, q.get("versions"))),
-        Q(os=q.get("os")),
+        Q(metric=kwargs["probe"]),
+        Q(channel=CHANNEL_IDS[kwargs["channel"]]),
+        Q(version__in=map(str, kwargs["versions"])),
+        Q(os=kwargs["os"] if "os" in kwargs else None),
```
robhudson: The same can be accomplished with `kwargs.get("os")`.  The `.get()` returns `None` by default if not found.

PR #191, opened 2020-01-14
(fixes #182) fix search results container position

### LINKED ISSUE #182: make the telemetry search box's position sticky
The search box, when the page body scrolls, scrolls w/ the body. This is because in calculating the position of the box we're also using `scrollY` from `<svelte:window />`

spasovski:
Seems to work fine locally - not sure if this is the right approach though.

### DIFF
--- a/src/app/sections/TelemetrySearchResults.svelte
+++ b/src/app/sections/TelemetrySearchResults.svelte
@@ -95,14 +95,15 @@ $: if (parentElement && windowWidth) {
   --list-border-color: gainsboro;
   --list-border: 1px solid var(--list-border-color);
   max-height: calc(100vh - var(--header-height) * 3);
-  border:1px solid var(--line-gray-02);
+  border: 1px solid var(--line-gray-02);
   background-color: white;
   max-width: calc(var(--increment) * 16);
   box-shadow: var(--depth-5l);
   border-bottom-right-radius: var(--border-radius-base);
   border-bottom-left-radius: var(--border-radius-base);
-  position: absolute;
-  left:0;top:0;
+  position: fixed;
+  left: 0;
+  top: var(--increment);
   overflow: hidden;
 }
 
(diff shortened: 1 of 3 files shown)


src/app/sections/TelemetrySearchResults.svelte
```diff
@@ -201,23 +201,22 @@ li {
 
 <svelte:window bind:innerWidth={windowWidth} on:keydown={handleKeypress} />
 
-<Portal>
+<Portal customClasses={['telemetry-search-results-container']}>
```
hamilton: I think this can actually be done without adding `customClasses` as a prop. adding to `.telemetry-results` a `position:fixed; top: var(--increment);` seems to achieve the same effect with minimal change to `Portal`.

hamilton (changes requested):
Thanks for the PR @spasovski! I left a note about a potentially smaller solution: just adding the `position:fixed; top: var(--increment)` directly to the `.telemetry-results` class (removing the `left` and `top` definitions like this (see bottom of declarations):

```css
.telemetry-results {
  --header-bg-color: var(--cool-gray-200);
  --list-border-color: gainsboro;
  --list-border: 1px solid var(--list-border-color);
  max-height: calc(100vh - var(--header-height) * 3);
  border:1px solid var(--line-gray-02);
  background-color: white;
  max-width: calc(var(--increment) * 16);
  box-shadow: var(--depth-5l);
  border-bottom-right-radius: var(--border-radius-base);
  border-bottom-left-radius: var(--border-radius-base);
  position: fixed;
  top: var(--increment);
  overflow: hidden;
}
```

This achieves what the `customClasses` part of `Portal` introduces without forcing that parent DOM element to have any style. I'm not opposed to that but I don't know if it is needed here, since it turns out `position: absolute` was simply the wrong declaration. Let me know what you think ~

src/app/sections/TelemetrySearchResults.svelte
spasovski: Yep I am all for added extensibility only when needed and this might be solved in the child DOM node indeed. Should be quick regardless.

hamilton:
lgtm now @spasovski – merging

PR #193, opened 2020-01-16
(fixes #184) add project footer

### LINKED ISSUE #184: add GLAM footer with github and legal links
good example a https://send.firefox.com

spasovski:
image


src/components/sections/Footer.svelte
```diff
@@ -0,0 +1,28 @@
...
+  grid-column: main;
+  display: flex;
+  justify-content: flex-end;
+  width: var(--main-content-width);
+  margin: var(--space-2x) auto;
+}
+.project-links {
+  list-style: none;
+  display: grid;
+  font-size: var(--text-015);
```
hamilton, line 12: can we also add a `font-weight: 500` here? This is a small nit, just a suggestion

src/components/sections/Footer.svelte
```diff
@@ -0,0 +1,28 @@
...
+  grid-gap: var(--space-2x);
+  grid-template-columns: auto auto auto;
+  padding: calc(var(--space-2x) * 0.7) var(--space-2x);
```
hamilton:
```suggestion
  padding: var(--space-base) var(--space-4x);
```

hamilton: Feel free to take or leave this suggestion:

(1) the 4x for the left/right helps it breath a little more against the curve of the body element. 
(2) `space-base` keeps the footer equal to the space from top of app  bar to the top of the body element

hamilton (commented):
looks good – two small ideas I thought of. Feel free to experiment to see what you think

hamilton:
On second thought, a few more things I'm thinking:

(1) the body element still goes to the bottom of the window. This can probably be relaxed so that if the body is not full height, the footer is still visible
(2) doing this would probably require reducing the height of the footer a bit so it matches the space b/t the body and the toolbar

PR #199, opened 2020-01-17
Fix #183: Fix inconsistent mouse events on search interface

### LINKED ISSUE #183: the search item click event works inconsistently – either it doesn't work at all or sometimes it selects another item
I had thought this bug was fixed, but it turns out it wasn't.

openjck:


src/app/sections/Search.svelte
```diff
@@ -16,27 +13,26 @@ let inputElement;
...
-    store.setField('searchIsActive', true);
+  store.setField('searchIsActive', true);
 }
 
-function turnOffSearch() {
-    setTimeout(() => {
-      store.setField('searchIsActive', false);
-    }, 100);
+function turnOffSearch(evt) {
+  console.log(evt);
```
spasovski: we can remove this

spasovski (changes requested):
I would clean up the console log and any editor-introduced whitespace that you don't want there. Tested just fine locally I think the main idea of removing the input blur event was a good call.

src/app/sections/Search.svelte
openjck: Whoops, thanks!

openjck:
Updated. Anything else before we merge?

src/app/sections/Search.svelte
```diff
@@ -135,12 +141,12 @@ async function onKeypress(event) {
       {/if}
     </div>
     <input 
+      type="search"
       aria-autocomplete="list"
       aria-controls="telemetry-search-results" 
       on:focus={turnOnSearch}
       bind:this={inputElement}
       placeholder="search for a telemetry probe"
-      on:blur={turnOffSearch}
```
spasovski, line 143: Seems we'll no longer need the `turnOffSearch()` function above since we set the store directly.

src/app/sections/Search.svelte
```diff
@@ -20,9 +20,7 @@ function turnOnSearch() {
 }
 
 function turnOffSearch() {
```
spasovski: Likely candidate for removal - or call this in the event handler below.

spasovski (changes requested):
r+wc

src/app/sections/Search.svelte
openjck: Yep, good call. Thank you!

PR #211, opened 2020-01-21
(fixes #118) add link to probe dictionary

### LINKED ISSUE #118: provide a link to the probe in Probe Dictionary
This will likely exist in the sidebar. Note that probe details links will need some formatting `https://probes.telemetry.mozilla.org/?view=detail&probeId=histogram%2FA11Y_CONSUMERS` for example.

spasovski:
image

### DIFF
--- a/src/app/sections/ProbeDetails.svelte
+++ b/src/app/sections/ProbeDetails.svelte
@@ -10,6 +10,7 @@ import { downloadString } from '../../utils/download';
 
 import Button from '../../components/Button.svelte';
 import StatusLabel from '../../components/StatusLabel.svelte';
+import ExternalLink from '../../components/icons/ExternalLink.svelte';
 
 const paneTransition = { x: 10, duration: 300 };
 
@@ -132,6 +133,18 @@ h2 {
     padding: var(--space-2x);
 }
 
+.more-info-link {
+    display: grid;
+    align-items: center;
+    grid-auto-flow: column;
+    grid-column-gap: var(--space-1h);
+    justify-content: end;
+    font-weight: 300;
+    padding-right: var(--space-1h);
+    margin: var(--space-2x) 0;
+    font-size: var(--text-01);
+}
+
 </style>
 
 <RightDrawer>
@@ -179,6 +192,9 @@ h2 {
                 <h2 class="detail__heading--01">description</h2>
                 <div class="probe-description helper-text--01">
                     {@html $store.probe.description}
+                    <a class="more-info-link" href={`https://probes.telemetry.mozilla.org/?view=detail&probeId=${$store.probe.apiName}`} target="_blank">
+                        more info <ExternalLink size=12 />
+                    </a>
                 </div>
             {/if}
         </div>
--- a/src/components/icons/ExternalLink.svelte
+++ b/src/components/icons/ExternalLink.svelte
@@ -1,0 +1,5 @@
+<script>
+export let size = 24;
+</script>
+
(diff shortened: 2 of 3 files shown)


src/components/icons/ExternalLink.svelte
```diff
@@ -0,0 +1,11 @@
+<script>
+export let size = 24;
+</script>
+
+<style>
+path {
```
hamilton: I tend to just add a `fill=currentColor` in the svg itself instead of using the style tag here so we don't introduce more classes. I'm not sure it matters in the end, and perhaps the best approach is a class on the svg that always sets `fill=currentColor`

src/app/sections/ProbeDetails.svelte
```diff
@@ -194,6 +199,13 @@ h2 {
             </div>
         </div>
         {/if}
+        <div class="drawer-section">
+            <div class="helper-text--01">
+                <a href={getProbeDictURL($store.probe.apiName)} target="_blank">
+                    open in Probe Dictionary <ExternalLink size=12 />
```
hamilton:
```suggestion
                    more info <ExternalLink size=12 />
```

hamilton: Not sure if this is the right copy, but worth considering what the right copy is here for the end-user

src/app/sections/ProbeDetails.svelte
```diff
@@ -194,6 +199,13 @@ h2 {
             </div>
         </div>
         {/if}
+        <div class="drawer-section">
+            <div class="helper-text--01">
+                <a href={getProbeDictURL($store.probe.apiName)} target="_blank">
```
hamilton: (1) the blue here feels very attention-drawing in the interface. I'm not sure this link needs the color, and the icon signals that it's probably a clickable link
(2) it'd be nice to get this more vertically aligned. Here's how I tend to do this (in this case, on the `a`):

```css
a {
  display: grid;
  grid-auto-flow: column;
  align-items: center;
  justify-content: end; <!-- or start, I dunno -->
  grid-column-gap: var(--space-1h);
}

```
(3) is this meant to be part of the description text? Seems like it belongs semantically near it, but that's just a hunch. It _does_ seem like it probaly belongs there vs. the `export` section.

src/app/sections/ProbeDetails.svelte
```diff
@@ -20,6 +21,10 @@ function probeIsSelected(probe) {
   return probe.name !== null && probe.name !== 'null';
 }
 
+function getProbeDictURL(apiName) {
```
hamilton: this is totally fine, but this can also be down below with 

```<a href="https://probes.telemetry.mozilla.org/?view=detail&probeId={$store.probe.apiName}">blahblahblah</a>```

I tend to do this since it removes a layer of indirection

**Review by hamilton — CHANGES_REQUESTED** [review]
Thanks for the PR. Had a couple of comments re: styling. Let me know what you think ~

File: src/app/sections/ProbeDetails.svelte (hunk 4 above)
Comment (spasovski) [reply src/app/sections/ProbeDetails.svelte]: The link logic had some branching before robhudson gave us the key in `apiName` so yes this is an old fragment.

File: src/app/sections/ProbeDetails.svelte [hunk 5]
```diff
@@ -194,6 +210,13 @@ h2 {
             </div>
         </div>
         {/if}
+        <!--div class="drawer-section">
```
hamilton: should this be deleted?

spasovski: indeed...i should review my own diffs better

PR #213, opened 2020-01-22
Fix #212: Update homepage after search

### LINKED ISSUE #212: Clicking on a probe from the homepage does not update view
## Steps to reproduce

1. Navigate to the homepage
2. Search for a probe
3. Click on it

## Expected result

The view updates, such that the user sees that probe.

## Actual result

The URL updates, but not the homepage. The homepage is still shown.

openjck:


hamilton (approved):
tested locally – lgtm. Thanks for the quick fix!

PR #215, opened 2020-01-22
(fixes #52) add generic docs link on probe type

### LINKED ISSUE #52: Add links from probe type to probe type docs on DTMO
The probe type text on a probe's detail page should link to the description of that probe type in the docs.telemetry.mozilla.org documentation.

spasovski:
Not fully sure about styling decisions but I had a chat with rmiller about where this should point to and it seems the link to measurement types is the best right now.


hamilton (changes requested):
visually looks great. I think `histogram` / `scalar` can map to [this](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/collection/histograms.html) and not docs.tmo, and in theory each `kind` can be mapped there as well

spasovski:
@rafrombrc what do you think of @hamilton 's suggestion to link to source-docs?

PR #238, opened 2020-02-14
Run Prettier on all supported files

openjck:
This commit adds Prettier to the project for formatting HTML, CSS,
JavaScript, JSON, and Markdown. A "format" script is also added to
format from the command line. Most editors can format on save, but I
still find this to be useful.

I don't agree with Prettier on everything. I wish it used trailing
commas and single quotes, for example. But I think consistency with
other projects is helpful and beginning down the road of discussing
different configuration options just seems like a waste of time.

I haven't had much luck with Svelte plugins, so this commit doesn't add
support for formatting Svelte files. We can revisit Svelte formatting
down the road.


openjck:
I haven't had any luck with Svelte plugins for Prettier, but we can revisit them down the road.

openjck:
Just a quick notice to anyone who drops by here: we can merge this whenever the time is right. This page will warn about conflicting files, but those are trivially easy to fix: just delete the latest commit and run `npm run format` again.

In other words, if you want to merge this but you're concerned about the conflicting files, just ping me and I'll take care of them within a couple of minutes.

openjck:
Prettier just released version 2.0, which will cause significant differences in formatted code. Before we merge this, let me upgrade to Prettier 2.0 and re-run formatting. Even with this it shouldn't take long at all to rebase this PR whenever we want to merge.

openjck:
Updated to use Prettier 2.0. This can be merged now or rebased when necessary.

jest.config.js
```diff
@@ -1,11 +1,8 @@
 module.exports = {
-  moduleFileExtensions: ['js'],
+  moduleFileExtensions: ["js"],
```
spasovski: Not a fan of doublequotes for JS strings personally and our previous projects have been with `'`. More important to be consistent though in either case.

spasovski:
I won't block on the singlequotes thing, esp if other projects have settled on it. Looks great otherwise r+

spasovski (approved):
Looks great thanks!

openjck:
Updated to use single quotes. Fine by me. That's what reviews are for! 😃

PR #245, opened 2020-02-25
add remaining color palettes

spasovski:
image


hamilton (changes requested):
This is a great start @spasovski. The new interpolations will be really useful. I think we should still go through and look at all the places where we have used `--digital-blue` in components and get those to match more closely the older color. In other words, let's commit to the palette and get the components that used the old colors to match those by changing the numbers. This will involve mostly changing color declarations used elsewhere (most of the new definitions are darker under 500 than they were in `master`). See the below GIF for an example on `GLAMMark.svelte`:

Kapture 2020-03-11 at 7 26 02

Issue #260, opened 2020-02-27
Fix failing window-functions.test.js tests


openjck:
This may be happening because udgl was moved to a module and therefore isn't being pre-processed by Babel any more.

openjck:
My forthcoming *routing.test.js* file is also untestable for the same reason. I've been messing around with the Jest config but I can't seem to solve this right now.

openjck:
Ah. Accidentally solved this while messing around with Babel for another reason. Not bad! This will be in my #261 PR.

PR #261, opened 2020-02-28
Fix #248: Add client-side routing with page.js

### LINKED ISSUE #248: Add client-side routing
I'm working on adding client-side routing using [page.js](https://github.com/visionmedia/page.js).

openjck:


src/components/Layout.svelte
```diff
@@ -35,8 +34,11 @@ $: if (visible) {
 
 onMount(() => { visible = true; });
 
+let width;
 </script>
 
+<svelte:window bind:innerWidth={width} />
```
hamilton: what is this for?

src/state/store.js
```diff
@@ -77,16 +78,8 @@ export function getFromQueryStringOrDefault(fieldKey, isMulti = false) {
 }
 
 const initialState = {
-  probe: {
```
hamilton, line 81: can you give some more detail about why you pulled probe out?

src/components/Layout.svelte
openjck: I'm not honestly sure. It came from *GraphicBody*. 😄

src/state/store.js
openjck: Sure! The short answer is that the current probe can be inferred from the path and synchronizing the URL (which is something like a global store itself) and the `store` object seemed fraught and unnecessary.

That said, this approach isn't perfect either. I'm very open to discussing this further.

For the time being, see `currentProbe` in *telemetry-search.js* which partly replaces this.

src/components/GLAM.svelte
```diff
@@ -0,0 +1,47 @@
...
+  page(
+    "/probe/:product/:name/table",
+    updateSection,
+    () => component = Table,
+  );
+
+  page.start();
+</script>
+
+{#if $store.token}
```
hamilton: we should file an issue here to add an `{:else} <Spinner/>` or something like that. Sometimes it takes a while for me to see anything and I think there's a bug

src/components/Layout.svelte
hamilton: it looks like it is unused here. Let's get rid of it

src/udgl/Details.svelte
```diff
@@ -1,6 +1,9 @@
 <script>
 import { fly } from 'svelte/transition';
```
hamilton: we should remove this from `src/udgl/` and put it in `src/components/` somewhere.

hamilton (changes requested):
This is looking great. The patterns feel pretty nice on the whole. A few bits of feedback:

- I notice when I click b/t the explore & table view, it re-renders the probe details. This probably doesn't need to happen and feels distracting. Could we pull out `ProbeDetails` so that it is part of the layout for `explore` and `table`?
- `http://localhost:5000/probe/` seems broken. Can we add a redirect for all the half-routes  back to `/` for now?
- we will definitely need some kind of 404 or something, since `localhost:5000/break-this` also doesn't work.

src/state/store.js
hamilton: I'm not averse to this, just curious. It seems like if we move to page-based views, we can have less stuff in a global store. Maybe each view has its own store / query string to capture what it needs. Seems like this is a step in that direction anyway.

hamilton:
Found a bug worth bringing up here. If I (1) go to the probe `http_altsvc_entries_per_header` then click on the GLAM logo, I get a `stop` error, suggesting that there's some problem w/ one of the stories (possibly the `currentProbe` store)

This error doesn't happen if the probe loads as intended (that is, no application error message that we'd expect)

Kapture 2020-02-28 at 15 50 48

hamilton:
Another issue worth considering: as it stands now, the back button still doesn't work with this PR. We can make that a separate issue to address later

src/components/Layout.svelte
openjck: 👍

src/udgl/Details.svelte
openjck: Unless I'm mistaken, it doesn't look like this or *DetailSection.svelte* is used at all. Do we plan to use them or are they artifacts of the past?

src/components/GLAM.svelte
openjck: Yeah, good catch. Thank you!

https://github.com/mozilla/glam/issues/272

openjck:
> Another issue worth considering: as it stands now, the back button still doesn't work with this PR. We can make that a separate issue to address later

Looking into this. I find that the back button does work except for the first time it's used. May have something to do with query params being added to the URL.

openjck:
> I notice when I click b/t the explore & table view, it re-renders the probe details. This probably doesn't need to happen and feels distracting. Could we pull out ProbeDetails so that it is part of the layout for explore and table?

✅

src/udgl/Details.svelte
hamilton: Good catch. Let's toss em.

openjck:
Pushed up my latest code. Most issues have been fixed, but the back button still doesn't work predictably and there are still sometimes `stop` errors. That said, I refactored this to make our lives easier down the road and fixed #260 and the circular *telemetry-search* dependency along the way, so I'll call it even.

I'll work on the last remaining issues tomorrow.

openjck:
> This is looking great. The patterns feel pretty nice on the whole. A few bits of feedback:
> 
>     * I notice when I click b/t the explore & table view, it re-renders the probe details. This probably doesn't need to happen and feels distracting. Could we pull out `ProbeDetails` so that it is part of the layout for `explore` and `table`?
> 
>     * `http://localhost:5000/probe/` seems broken. Can we add a redirect for all the half-routes  back to `/` for now?
> 
>     * we will definitely need some kind of 404 or something, since `localhost:5000/break-this` also doesn't work.

These are all fixed.

openjck:
I'm still occasionally getting strange behavior from the back button, including the `stop()` error, but it works as expected much of the time. I'll continue looking into it, but I don't think the fix will be too dramatic, so I think it's safe to give this another review.

openjck:
Ready for review. The only issue is that the back button sometimes behaves strangely. I'm not sure what that's about, but I'd be content to split that off into another GitHub issue since the back button doesn't work on master right now, either.

openjck:
Actually, the back button may be working pretty well after all. I might have been seeing another issue (clicking *Table* when no data is available raises a `stop` error and breaks the back button). Let me know how it works for you.

src/state/store.js
```diff
@@ -77,17 +77,43 @@ export function getFromQueryStringOrDefault(fieldKey, isMulti = false) {
...
+  const pathParts = url.pathname.split('/').filter((p) => p !== '');
+
+  const value = {
+    raw: absoluteURL,
+    path: {
+      section: pathParts[0],
+    },
+  };
+
+  if (value.path.section === 'probe') {
```
hamilton: so if I add a new `section`, where are all the places I'm going to have to update this? Here, and in the `GLAM.svelte` file? Is there a way we can consolidate these somehow? We will definitely need to figure out how, down the line, we add more routes + ways of setting things.

src/utils/routing.js
```diff
@@ -0,0 +1,48 @@
...
+  window.addEventListener('popstate', (evt) => {
+    if (evt && evt.state) {
+      updateStoreURL(makeAbsolute(evt.state.path));
+    }
+  });
+}
+
+export function probeURL({ product = 'firefox', name, view = 'explore' } = {}, preserveQuery = true, preserveHash = true) {
+  const currentURL = new URL(window.location.href);
+  let nextURL = `/probe/${product}/${name}/${view}`;
```
hamilton: After some thinking, I would opt to reverse this a bit: I'd like to see `/${product}/probe/${name}/${view}` instead, I think, since I think down the line products might have other sections (like `probe`, not like the `view` route here), but not every view is necessarily going to be available for each product.

src/utils/routing.js
```diff
@@ -0,0 +1,48 @@
...
+
+  // Back/Forward buttons
+  window.addEventListener('popstate', (evt) => {
+    if (evt && evt.state) {
+      updateStoreURL(makeAbsolute(evt.state.path));
+    }
+  });
+}
+
+export function probeURL({ product = 'firefox', name, view = 'explore' } = {}, preserveQuery = true, preserveHash = true) {
```
hamilton: this is good, but it does seem like this is a third place I'll have to update / add new url functions when I add a new route / view (first is in `GLAM.svelte`, the second is in `buildStoreURLValue`). Is there a way we can consolidate this so future maintainers that aren't you and me can jump into this more easily? I'm looking for patterns that feel really easy and effortless that we can bring into other apps. This feels like it could lead to bugs & confusion as it is now.

src/utils/routing.js
```diff
@@ -0,0 +1,48 @@
...
+  const currentURL = new URL(window.location.href);
+  let nextURL = `/probe/${product}/${name}/${view}`;
+  if (preserveQuery) nextURL += currentURL.search;
+  if (preserveHash) nextURL += currentURL.hash;
+  return nextURL;
```
hamilton: Additional suggestion: we could have a section key (probe) and generalize this function to be any GLAM URL route.

src/utils/routing.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function navigateToProbe(probeDetails, preserveSearch = true, preserveHash = true) {
```
hamilton: if you take the suggestion above and generalize `probeURL` to be just `glamURL` or something, then this can just be `navigate` possibly.

hamilton (changes requested):
This is shaping up real nicely, @openjck! A few comments & questions for ya:

(1) This might be a bug on `master` (I don't have a ton of bandwidth to check right now) but if I reload on a probe but click on the GLAM logo before it loads fully, I get a dreaded `stop` error. If the bandwidth is running thin to fix it on this PR, let's file a bug for fixing this later (could be something simple like needing to unsubscribe somewhere) so we can slot in the fix later. See the GIF:

Kapture 2020-03-06 at 16 22 56

(2) Seems like we might be able to consolidate a number of the routing configuration stuff into the same place, so if someone not totally familiar with this has to implement a new view or tweak anything, it won't take them forever to figure things out. As it is now, it seems like I might have to check in a handful of places to implement a new view.

(3) I like the pattern you came up with with `$store.url`, but is it needed? I could use a walkthrough. It seems I could just pull out `ctx.params` from the callback in `page` over in `GLAM.svelte` and potentially just pass those as props into the route component directly. This would probably also require refactoring `Layout.svelte` into `GLAM.svelte` directly so we don't have two places to deal with this logic. If I had something like this:

```
page('/:product/probe/:name/:view', (ctx, next) => {... })
```

I can take a look at `ctx.params` to pull out `product`, `name`, and `view` directly, and then pass them down into a component defined in the template. Or we could have a store that we use, called `route`, that gets passed down into the children w/ all the route params. I don't have a preference, but as it is right now it feels a bit indirect, since these params are needed only in the routing logic and not really anywhere else (pageload can be handled by looking at hte query string, and I think search / clicking on a probe works w/o `$store.url` right?

(4) The back button works for the changes in the route, but it doesn't seem to work on changes to the query string. I don't think `registerStoreURLUpdaters` fires on query string updates (which happens when you click on body items like buttons). Is it supposed to? I don't think it _needs_ to, necessarily, just wanted to bring this up.

(5) It seems to reload the page if I'm going from loaded probe -> splash page using the back button (that is, start on the splash, click to a probe, then click back button). I haven't noticed any other obvious weirdness, but I could try more after you go through these comments.

---

At any rate, let me know what you think!

src/utils/routing.js
openjck: Thanks! I wanted to add `buildStoreURLValue` to *store.js* to avoid a circular dependency, but maybe there's a better approach here.

openjck:
Thanks for the detailed feedback. I'd rather take the time to get this right, even though this is getting pretty far behind my original estimate, than merge something that's problematic and spend even more time fixing it.

I have some questions which I'd like to record right now so that I don't forget them, but we can talk about them next week. I'll add questions to this comment as more come to mind.

1. You mention that you'd like to consolidate things. I agree. I was thinking of adding a *src/routing* directory which contains three items: a *router.svelte* file (the main page configuration), a *utils.js* file (`navigateToProbe` and the like, as needed), and a *pages* directory (one Svelte component for each page). Some minor details will change, but is this generally what you had in mind or am I totally off-base?

hamilton:
@openjck yeah something like that seems about right to me. Regarding consolidation, I'm mostly thinking about the developer / maintainer experience – that should help clarify things. Put yourself in another john's perspective – if a john clone were to approach this code base for the first time and he were tasked with adding a new route, how would you organize this to make it as frictionless an experience as possible for him? I understand the irony that GLAM is not in some parts developer-friendly, but we're doing our best :) . However you choose to make it easy easy to add / amend works for me, and preferably editing only one or two files.

I don't think what you have is that far off, and certainly we can work with it if we don't have more time to invest in this – I leave that to you and Rob to figure out. The only thing that seems critical to get right is the exact routing in the URL, which is a very easy change. The rest is invisible to the end user.

src/utils/routing.js
hamilton: Depending on the remaining work appetite for Data Tools on this, this suggestion would be the one thing we should agree upon no matter what since it's technically the only bit of UX this PR introduces.

openjck:
Updated and ready for another review. 😄

src/routing/Router.svelte
```diff
@@ -0,0 +1,95 @@
+<script context="module">
+  // We could grab the value of currentQuery here rather than requiring that it
```
hamilton, line 2: passing it as an argument is probably the better move, because it allows `getPath` to be updated reactively in templates (as in `GLAMMark`) when the function args update

hamilton: If you wanted to get an up-to-date store value, you could use `get` I guess. But yeah this won't update the hrefs elsewhere when the querystring changes. That's an important part I think, since this appends the current query to the rest of the path

src/routing/Router.svelte
```diff
@@ -0,0 +1,95 @@
...
+        name,
+        view,
+      });
+
+      store.setField('showingProbe', item === 'probe' ? true : false);
+    };
+  }
+
+  page('/', useComponent(Home));
+  page('/:product/:item/:name/explore', useComponent(ProbeExplore, 'explore'));
```
hamilton: minor nit – would you mind if we change `:item` to `:section` here + in the store? Item feels vague, and I think we're likely to add other sections for products in the future

src/components/controls/ProbeViewControl.svelte
```diff
@@ -23,15 +24,15 @@ export let transformed = options.map((opt) => ({
 }));
 </script>
 
-{#if $store.appView === 'PROBE'}
+{#if $store.showingProbe}
```
hamilton: why use `showingProbe` when we have `$store.route.item`?

src/routing/Router.svelte
```diff
@@ -0,0 +1,95 @@
+<script context="module">
+  // We could grab the value of currentQuery here rather than requiring that it
+  // be passed as an argument, but it's unclear how to get an up-to-date store
+  // value from within a module.
+  export function getPath(path, query, preserveQuery = true) {
```
hamilton: What is the use case for `preserveQuery` ? It seems if I don't want to preserve the query, I just wouldn't provide `query`, right?

src/routing/Router.svelte
hamilton: Also, I suspect we will have to refactor this if we have other `section`s that have a different routing structure. The URL structure will end up being the same to the end-user. This goes against my previous recommendation :) let's go with this for now, but be ready to change it later

hamilton (dismissed):
Looks great – thanks for making all those changes. This is definitely more readable, though I bet we'll have to refactor `Router.svelte` down the line to be a bit more flexible. For now, this works great.

I had a few minor nits. Feel free to merge when you feel comfortable w/ those.

src/routing/Router.svelte
openjck: > If you wanted to get an up-to-date store value, you could use `get` I guess. But yeah this won't update the hrefs elsewhere when the querystring changes. That's an important part I think, since this appends the current query to the rest of the path

Exactly. That's the part I couldn't figure out. Argument it is!

src/routing/Router.svelte
openjck: 👍

src/routing/Router.svelte
openjck: Good call. This was an artifact of me trying to `get` the query within the function's implementation. I'll refactor this.

PR #266, opened 2020-02-28
add error condition for unlikely table duplicate data

spasovski:
image

### DIFF
--- a/src/components/regions/GraphicBody.svelte
+++ b/src/components/regions/GraphicBody.svelte
@@ -61,7 +61,6 @@ function handleBodySelectors(event) {
     store.setField('activeBuckets', selection);
   }
 }
-
 </script>
 
 <style>
@@ -116,7 +115,7 @@ h2 span {
 } */
 
 .graphic-body__content {
-  box-sizing: border-box; 
+  box-sizing: border-box;
   grid-area: content-body;
   min-height: var(--height);
   background-color: white;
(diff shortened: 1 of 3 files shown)


src/components/regions/GraphicBody.svelte
```diff
@@ -178,49 +177,49 @@ h2 span {
                   <Spinner size={48} color={'var(--cool-gray-400)'} />
               </div>
               {:then data}
-                  {#if $store.probeView === 'explore'}
-                    <div in:fade class="graphic-body__content">
-                        {#if $temporaryViewTypeStore === 'categorical'}
```
hamilton, line 183: for your sake, if this is all just formatting fixes, I would undo this change since it will collide w/ john's PR

src/components/table/ProbeTableView.svelte
```diff
@@ -40,53 +48,57 @@ let currentKey = probeKeys[0] || undefined;
...
-      {#if aggregationTypes && aggregationTypes.length > 1}
-      <div class=body-control-set>
-        <label class=body-control-set--label>Metric Type</label>
-        <AggregationTypeSelector 
-          aggregationTypes={aggregationTypes}
-          bind:currentAggregation={currentAggregation}
-        />
-      </div>
-      {/if}
+  {#if isDuplicateDataFound}
```
hamilton, line 53: Can we pass on a generalized error here instead? `isDuplicateDataFound` is definitely a weird edge case that shouldn't occur very often if at all, and there may be other errors caught / thrown here.

hamilton (commented):
Thanks for taking this on @spasovski. I'm wondering if it might be a safer approach to catch _any_ error that occurs here instead of special-casing this one-off that is not likely to occur again in the future. Let me know what you think. Apologies this took a while – I had pending comments that I had not submitted, apparently!

PR #268, opened 2020-03-02
add process probe filtering

spasovski:
image


hamilton (commented):
I think this is a good start – thanks @spasovski. We don't do anything here to return a useful error message when someone switches process on a selected probe but it does not exist. The error message is a "nice" one but it is not clear _why_ there is no data. We should file an issue to catch this, since it might create confusion as it is now. This PR is good though – let's merge it.

PR #282, opened 2020-03-05
adds client-side response validation

### LINKED ISSUE #277: create validators for probe data
A large class of bugs in the interface come from a malformed response. We need a validation step to throw a human-readable error if one is to be had. This should look like something like this:

```
validate(data, hasResponse, noDuplicates, hasRightMetadata, hasRightKeys)
``` 

And so on.

This allows us to:
- write tests for the validation
- move out a huge amount of error-catching display-breaking logic from components into pure JS land, which is more tractable on the whole
- improve maintainability

### LINKED ISSUE #280: very low-count probes get filtered out unnecessarily
This is an easy fix, so I'll take it on. The function `topKBuildsPerDay`, which is was a stopgap for having far too many nearly-empty builds, was filtering out all builds w < 10 counts for probes. Some probes have very low counts, so this creates a weird data validation problem.

### LINKED ISSUE #283: add a generalized 500 error
Not all errors will be caught. But we can catch 500 errors by throwing in `api.js` when the response status is in the 500 range.

hamilton:
closes #280, closes #277, closes #283

There are many classes of bugs in GLAM that ultimately boil down to something malformed in the data. We should be checking the response for various cases upfront and throwing in the data transformation step. This will thankfully bubble up in a nice, human-readable way in the frontend.

Adding validation early and in pure JS will also allow us to write comprehensive tests for anomalies and craft nice error messages for our users.

This isn't meant to be a comprehensive set of validators; it should, however, give maintainers enough of a blueprint to go in and implement new validating functions. It also gives us a rubric for where we should throw. If there is a problem with the API response, we throw early. If there is a problem with anything downstream of the API response, we can throw there instead.

This PR addresses the following:
- adds custom validators, as described above
- fixes bug in the top k build ids workflow
- adds a generalized 500 error so we can gracefully guide the user toward reporting on the GLAM channel

### DIFF
--- a/src/state/store.js
+++ b/src/state/store.js
@@ -126,7 +126,7 @@ export const resetFilters = () => {
   store.setField('channel', getDefaultFieldValue('channel'));
   store.setField('os', getDefaultFieldValue('os'));
   store.setField('aggregationLevel', getDefaultFieldValue('aggregationLevel'));
-  store.setField('process', getDefaultFieldValue('process'))
+  store.setField('process', getDefaultFieldValue('process'));
 };
 
 export const searchResults = derived(
@@ -273,7 +273,7 @@ export function fetchDataForGLAM(params) {
       validate(
         payload,
         (p) => noResponse(p, probe.active),
-        (p) => noDuplicates(p.response, aggregationLevel),
+        (p) => noDuplicates(p, aggregationLevel),
       );
 
       const probeType = probe.type;
--- a/src/state/store.js
+++ b/src/state/store.js
@@ -12,6 +12,7 @@ import CONFIG from '../config.json';
 
 import { byKeyAndAggregation, getProbeViewType } from '../utils/probe-utils';
 
+import { validate, noDuplicates, noResponse } from '../utils/data-validation';
 
 export function getField(fieldKey) {
   return CONFIG.fields[fieldKey];
@@ -203,7 +204,6 @@ function paramsAreValid(params) {
 
 export const datasetResponse = (level, key, data) => ({ level, key, data });
 
-
 // FIXME: let's remove this function. It's almost comically redundant.
 export function responseToData(data, probeClass = 'quantile', probeType, aggregationMethod = 'build_id') {
   return byKeyAndAggregation(data, probeClass, aggregationMethod, { probeType }, { removeZeroes: probeType === 'histogram-enumerated' });
@@ -270,11 +270,11 @@ export function fetchDataForGLAM(params) {
       const { probe } = st;
       const { aggregationLevel } = st;
 
-      if (!('response' in payload)) {
-        const er = new Error('The data for this probe is unavailable.');
-        if (!probe.active) er.moreInformation = 'This probe appears to be inactive, so it\'s possible we don\'t have data for it.';
-        throw er;
-      }
+      validate(
+        payload,
+        (p) => noResponse(p, probe.active),
+        (p) => noDuplicates(p.response, aggregationLevel),
+      );
 
       const probeType = probe.type;
       const probeKind = probe.kind;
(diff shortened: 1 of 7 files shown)


src/utils/probe-utils.js
```diff
@@ -149,7 +149,7 @@ export function zipByAggregationType(payload) {
 export function topKBuildsPerDay(dataset, k = 2) {
   const byBuildID = groupBy(dataset, 'label', formatBuildIDToOnlyDate);
   const topK = Object.entries(byBuildID).map(([_, matches]) => {
-    const out = matches.filter((m) => m.audienceSize > 10);
+    const out = matches;
```
hamilton, line 152: here is an example of a downstream bug. The filtering here actually created an issue where data was missing because we were filtering out (for some reason, likely historical ones) data points with a small audience size.

src/state/store.js
```diff
@@ -125,7 +126,7 @@ export const resetFilters = () => {
   store.setField('channel', getDefaultFieldValue('channel'));
   store.setField('os', getDefaultFieldValue('os'));
   store.setField('aggregationLevel', getDefaultFieldValue('aggregationLevel'));
-  store.setField('process', getDefaultFieldValue('process'))
+  store.setField('process', getDefaultFieldValue('process'));
```
openjck, line 129: Fixes an eslint error! 👍

tests/validators.test.js
```diff
@@ -0,0 +1,40 @@
...
+  it('passes if there is a response key', () => {
+    expect(() => noResponse(noResponseCases.hasResponse)).not.toThrow();
+    expect(() => noResponse(noResponseCases.noResponse)).toThrow();
+  });
+});
+
+describe('noDuplicates', () => {
+  it('passes or rejects depending on if there are duplicate keys (build_ids, versions)', () => {
+    expect(() => noDuplicates(noDupsData.noDups, 'build_id')).not.toThrow();
+    expect(() => noDuplicates(noDupsData.dups, 'build_id')).toThrow();
```
openjck: Nitpick, but the name `noDupsData.dups` tripped me up for a second. Maybe `dupsData` or `dupsTestData`?

tests/validators.test.js
```diff
@@ -0,0 +1,40 @@
...
+    { metadata: { build_id: 'abcd' } },
+    { metadata: { build_id: 'abcd' } },
+  ],
+};
+
+
+describe('noResponse', () => {
+  it('passes if there is a response key', () => {
+    expect(() => noResponse(noResponseCases.hasResponse)).not.toThrow();
+    expect(() => noResponse(noResponseCases.noResponse)).toThrow();
```
openjck: Same here. Maybe `responseData` or `responseTestData`?

src/utils/data-validation.js
```diff
@@ -0,0 +1,21 @@
...
+  const uniques = new Set(allBuildIDs);
+  if (allBuildIDs.length !== uniques.size) {
+    throw new Error(`Duplicate ${aggregationMethod === 'build_id' ? 'Build IDs' : 'Versions'} found.`);
+  }
+};
+
+export const noResponse = (payload, probeIsActive) => {
+  if (!('response' in payload)) {
+    const er = new Error('The data for this probe is unavailable.');
+    if (!probeIsActive) er.moreInformation = 'This probe appears to be inactive, so it\'s possible we don\'t have data for it.';
```
openjck, line 14: Another nitpick (maybe Prettier would have an opinion on this?) but our eslint config allows us to use double outer quotes if there are single quotes in the string, as an alternative to escaping.

openjck:
I like this. Looks good to me other than a couple of nitpicks. 👍

openjck (approved):

PR #293, opened 2020-03-06
(fixes #201) ellipsis long category text overflow

### LINKED ISSUE #201: long string key values cause unfortunate overflows
Kapture 2020-01-17 at 17 08 20

spasovski:
clean version of the previous PR


hamilton (approved):
Thanks @spasovski – this looks good. There are a couple of various cases where the total length of the table cuts into the padding a bit, but we can address that another time.

PR #295, opened 2020-03-10
(fixes #292) wrap probe details bug numbers

### LINKED ISSUE #292: bug numbers 'css is awesome' overflow
search for http_traffic_analysis_3

image

image

spasovski:
image


spasovski:
likely just needs a spot-check, pretty trivial fix

openjck:
Looks great! This also avoids the issue where the last row is spaced strangely, for example when `justify-content: space-between` is used and there's fewer items in the last row. That always trips me up with flexbox.

PR #296, opened 2020-03-10
(fixes #278) hide export JSON if data error occurs

### LINKED ISSUE #278: hide export when there is a data-level error
We could stand to hide the export functionality instead of showing this error message.

Screen Shot 2020-03-04 at 3 08 38 PM

spasovski:
I am not actually sure if this is the right approach though it certainly gets the job done. It should be fine since the main content area would have displayed a more detailed/prominent message anyway.

image


hamilton (approved):
Yep, this is what I have in mind – thanks @spasovski

PR #299, opened 2020-03-11
Fix bug with authentication in client-side routing

openjck:


openjck:
Got an r+ from Hamilton over Slack. 👍

PR #300, opened 2020-03-11
Fix bug where back button causes refresh 

openjck:


openjck:
#300 🎳

hamilton (approved):
Solid. Thanks @openjck

PR #306, opened 2020-03-11
Fix #272: Show layout and spinner during auth

### LINKED ISSUE #272: Show a spinner while authentication is being processed
Inspired by [this comment](https://github.com/mozilla/glam/pull/261#discussion_r385954573). We should show a spinner while authentication is being processed. I've seen the same behavior that Hamilton is seeing. When auth0 redirects back to GLAM, there can be a few moments where `code` and `state` are in the URL before `handleRedirectCallback` is run.

openjck:


src/state/store.js
```diff
@@ -251,7 +255,7 @@ export function isCategorical(probeType, probeKind) {
 }
 
 export function fetchDataForGLAM(params) {
-  return getProbeData(params, store.getState().token).then(
+  return getProbeData(params, get(store).auth.token).then(
```
hamilton: hmm, re: your comment in Slack, no strong preference, but generally it is probably wiser to use `store.getState()` in case we have to somewhere else change the implementation details in some way.

hamilton (changes requested):
This is great – thanks @openjck. I would use `store.getState()` where we can, generally, though I'm sure I've been all-over-the-place with this and should do better. Feel free to merge once that change is in ~

PR #307, opened 2020-03-12
fixes #265 - scroll long probe key floatingmenus

### LINKED ISSUE #265: huge keysets overflow the menu
a great example of this is `webext_storage_local_idb_set_ms_by_addonid`.

Some probes just have way too many keys. We should make the element within the floating menu (1) not go below the bottom of the page, and (2) scrollable.

Screen Shot 2020-02-28 at 3 32 38 PM

spasovski:


src/components/controls/ProbeKeySelector.svelte
```diff
@@ -8,24 +8,24 @@
   export let options;
   export let currentKey;
   export let active;
-  
+
```
openjck, line 11: Not a big deal, but for the time being, it may make sense to turn off the editor settings that are removing trailing spaces and adding newlines at the ends of files, just to make the diffs a little bit less noisy.

src/udgl/menu/FloatingMenu.svelte
```diff
@@ -73,7 +77,8 @@ $: if (element && parent) {
   <div class=bound-menu bind:this={element} style="
     left: {leftPlacement}px;
     top: {topPlacement}px;
+    max-height: calc(100vh - {topPlacement}px - var(--space-6x));
```
openjck, line 80: Nice!

src/udgl/menu/MenuList.svelte
```diff
@@ -58,7 +58,6 @@ ul {
   margin:0;
   padding:0;
   background-color: white;
-  border: 1px solid var(--line-gray-01);
```
openjck, line 61: Are there no other `ul`s which will be affected by this?

src/udgl/menu/FloatingMenu.svelte
hamilton: I would like to see the `var(--space-6x)` be codified somewhere else, maybe as `--screen-buffer` or something like that.

hamilton: or `--screen-padding`?

src/udgl/menu/MenuList.svelte
spasovski: Really good catch! Though oddly enough I didn't find visual inconsistencies in the top-right menus for example (which use this) but still this needs to be reverted.

PR #312, opened 2020-03-12
package.json cleanup

openjck:


hamilton (approved):
haven't tested this, but these are all sensible changes. I might also suggest we set the version to "0.1.0" until we have tagged release deployments, but consider that a take-or-leave.

openjck:
Updated 😃

Issue #237, opened 2020-02-14
Remove unneeded / unused commented code


openjck:
I would be happy to help with this and could propose any changes in pull requests.

I'm sometimes a bit confused when encountering commented code. It's hard to know if the code is commented only temporarily or if it's genuinely not likely to be needed in the future.

openjck:
The [eslint-plugin-extra-rules](https://www.npmjs.com/package/eslint-plugin-extra-rules) package has a *no-commented-out-code* rule which could help. It's not perfect and we may not want to officially add it to our eslint config (it ignores some commented code and marks some legitimate comments as code) but it may be somewhat helpful locally, as long as we don't commit it as an addition to our official eslint config.

PR #313, opened 2020-03-12
Issue #237: rm unneeded commented code from state/

openjck:


openjck:
It seems like all of this commented code is unnecessary, but correct me if I'm wrong.

hamilton (approved):

PR #318, opened 2020-03-12
Fix #316: Authenticate before fetching

### LINKED ISSUE #316: probe shows "data not available" after probe detail page refresh
step 1: search for any probe
step 2: after the probe loads hit refresh

result: data not available

expected: same probe details as when probe was loaded from search should show up

openjck:


spasovski:
Looks good thanks!

Issue #251, opened 2020-02-27
Fix eslint warnings and errors


openjck:
It may be easiest to do this in batches rather than committing everything all at once.

openjck:
After this is done, we should probably run eslint automatically some way, maybe in CI or in the dev server, so that we don't accidentally add more eslint warnings/errors in the future.

robhudson:
There are currenlty zero lint errors and linting is now enabled in CI runs.

PR #324, opened 2020-03-14
Issue #251: Address most routing/ eslint problems

openjck:


src/routing/pages/probe/Explore.svelte
```diff
@@ -10,13 +10,12 @@
...
   function handleBodySelectors(event) {
     const { selection, type } = event.detail;
-    if (type === "percentiles") store.setField("visiblePercentiles", selection);
-    if (type === "timeHorizon") store.setField("timeHorizon", selection);
-    if (type === "metricType")
-      store.setField("proportionMetricType", selection);
-    if (type === "activeBuckets") {
-      store.setField("activeBuckets", selection);
-    }
+    const renames = {
```
spasovski, line 13: `renames` is probably not the best name but I really can't think of something better. `metricTypeToStoreMapping` seems too verbose for something only used once here

spasovski (approved):
Looks good thanks!

PR #328, opened 2020-03-16
fixes #311 - deal with top axis labels with 1 point

### LINKED ISSUE #311: duplicate key issue
Leaving this for later. The following probe doesn't work, locally:

`http_page_revalidation`

spasovski:
Couldn't find more cases where this breaks locally.


spasovski:
I don't think this is needed at all after the refactor as part of the total clients work. Tested this against the original issue and it works fine locally and on dev.

PR #330, opened 2020-03-17
Align ProbeViewControl with left edge of page

openjck:


openjck:
Actually, I don't love the way this looks. I see now that the point was to align the left edge of the button with the left edge of the page text.

PR #332, opened 2020-03-18
fixes #321 - move probe details title

### LINKED ISSUE #321: Add breadcrumb to 'data for this probe is unavailable' page
When looking at a probe's data, there is an 'explore/<probe_token>' breadcrumb indicator at the top of the main content area. This doesn't show up on the 'data unavailable' page that shows upwhen the probe has no data. This means there's no place on the page that shows the token id of the probe the user is trying to view.

spasovski:
Surprisingly not simple.

### DIFF
--- a/src/components/errors/DataError.svelte
+++ b/src/components/errors/DataError.svelte
@@ -7,9 +7,7 @@ export let moreInformation;
 
 <style>
 .data-error-msg {
-  /* background-color: var(--cool-gray-050); */
   border-radius: var(--space-1h);
-
   padding: var(--space-4x);
   display: grid;
   align-items: center;
@@ -18,13 +16,12 @@ export let moreInformation;
   height: 300px;
   --error-msg-width: 400px;
   --error-msg-color: var(--cool-gray-500);
-  margin-top: var(--space-8x);
+  margin-top: var(--space-16x);
   margin-bottom: var(--space-2x);
 }
 
 .data-error-msg__bg {
   background: radial-gradient(var(--cool-gray-100), var(--cool-gray-100));
-  /* box-shadow: inset 0px 0px 10px rgba(0,0,0,.1); */
   width: 200px;
   height: 200px;
   padding: var(--space-4x);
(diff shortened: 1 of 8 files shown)


openjck:
**IF** we care about eslint, which may be an open question (you made a good point that it's less important with compiled languages), I wanted to note that this brings our eslint problem count up from 228 to 230, with the following being new:

```
[...]/glam/src/components/regions/ProbeTitle.svelte
  1:9  error  Too many blank lines at the beginning of file. Max of 1 allowed  no-multiple-empty-lines

[...]/glam/src/routing/pages/probe/Explore.svelte
  8:74  error  Trailing spaces not allowed  no-trailing-spaces
```

src/components/errors/DataError.svelte
```diff
@@ -7,9 +7,7 @@ export let moreInformation;
 
 <style>
 .data-error-msg {
-  /* background-color: var(--cool-gray-050); */
```
openjck, line 10: 👍

src/components/errors/DataError.svelte
```diff
@@ -18,13 +16,12 @@ export let moreInformation;
...
   --error-msg-width: 400px;
   --error-msg-color: var(--cool-gray-500);
-  margin-top: var(--space-8x);
+  margin-top: var(--space-16x);
   margin-bottom: var(--space-2x);
 }
 
 .data-error-msg__bg {
   background: radial-gradient(var(--cool-gray-100), var(--cool-gray-100));
-  /* box-shadow: inset 0px 0px 10px rgba(0,0,0,.1); */
```
openjck, line 27: 👍

src/routing/pages/probe/Explore.svelte
```diff
@@ -21,6 +22,7 @@
 
 <Probe let:data let:probeType>
   <div in:fade class="graphic-body__content">
+    <ProbeTitle view={$store.route.view} probeName={$store.probe.name} />
```
openjck, line 25: Do we think there will be cases where `view` is equal to something other than `$store.route.view`? If not, would it make sense to consider not taking this as a prop at all, but instead accessing `$store.route.view` from the `ProbeTitle` component?

openjck (changes requested):
Otherwise this is good. I like it!

src/components/explore/ProportionExplorerView.svelte
```diff
@@ -91,8 +88,8 @@ h2 {
       {/if}
     </div>
 
-    <div class=body-control-set>
-      <label class=body-control-set--label>Categories</label>
+    <div class="body-control-set">
```
hamilton, line 91: I'm seeing a lot of changes like this. Should we make this a lint rule so it is enforced? I have no preference but I typically don't put quotes in out of habit

PR #335, opened 2020-03-19
fixes #320 - change default process filter

### LINKED ISSUE #320: Switch 'process' default to 'content'
The process selector is currently defaulting to parent, which supports far fewer probes than content, we should switch the default so beginner users don't see the 'data for this probe is unavailable' page so often.

spasovski:
fixes #320

image

### DIFF
--- a/src/state/store.js
+++ b/src/state/store.js
@@ -14,6 +14,9 @@ import { byKeyAndAggregation, getProbeViewType } from '../utils/probe-utils';
 
 import { validate, noDuplicates, noResponse } from '../utils/data-validation';
 
+// TODO: move this to the new config.js when 'product' is added.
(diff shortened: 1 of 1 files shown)


src/state/store.js
```diff
@@ -14,6 +14,9 @@ import { byKeyAndAggregation, getProbeViewType } from '../utils/probe-utils';
 
 import { validate, noDuplicates, noResponse } from '../utils/data-validation';
 
+// TODO: move this to the new config.js when 'product' is added.
```
hamilton, line 17: always a fan for adding notes like this – thanks!

hamilton (approved):
Thanks @spasovski – looks good to me. This doesn't fully fix the missing probe issue I'm seeing locally, but this might be due to lack of data (I get a 404 in the server logs)

PR #340, opened 2020-03-20
fixes #336 - send process param to random probes

### LINKED ISSUE #336: the random probe API should take the same values as `/api/v1/data`
At the moment, the probes we are showing have hard-coded defaults. As such, if you click on one of htem, you then send a request to `/api/v1/data` that might not necessarily be there, since the params we pass into `/api/v1/data` might not be applicable.

The solution should be straightforward – we need to make sure the random probe API takes in the exact same params as `/api/v1/data`

spasovski:
This should be the right approach and I no longer get "missing data" probes on the Home view.

### DIFF
--- a/glam/tests/test_api.py
+++ b/glam/tests/test_api.py
@@ -36,7 +36,7 @@ class TestRandomProbesApi:
             "version": "72",
             "os": "Windows",
             "build_id": "*",
-            "process": 0,
+            "process": constants.PROCESS_CONTENT,
             "agg_type": constants.AGGREGATION_HISTOGRAM,
             "metric": name,
             "metric_key": "",
--- a/src/routing/pages/Home.svelte
+++ b/src/routing/pages/Home.svelte
@@ -9,6 +9,9 @@
   import { store, currentQuery } from '../../state/store';
   import { getRandomProbes } from '../../state/api';
   import { getPath } from '../Router.svelte';
+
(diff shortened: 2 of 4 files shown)


glam/tests/test_api.py
```diff
@@ -36,7 +36,7 @@ def _create_aggregation(self, name="gc_ms", data=None, multiplier=1.0):
             "version": "72",
             "os": "Windows",
             "build_id": "*",
-            "process": 0,
+            "process": 2,
```
robhudson: Can you use the constant here? `constants.PROCESS_CONTENT`

robhudson (approved):
r+wc

src/routing/pages/Home.svelte
```diff
@@ -9,6 +9,8 @@
   import { store, currentQuery } from '../../state/store';
   import { getRandomProbes } from '../../state/api';
   import { getPath } from '../Router.svelte';
+
+  const NUMBER_OF_RANDOM_PROBES = 9;  // TODO: add this to the upcoming config.js
```
openjck: New eslint error:

```
  13:37  error  Multiple spaces found before '// TODO: add t...'  no-multi-spaces
```

openjck: 1337

openjck:
Otherwise looks good to me!

PR #353, opened 2020-03-24
tweaks to summary table

### LINKED ISSUE #345: the summary table does not use the re-calculated percentiles
Noticed this during a meeting w/ data science today. One uses `transformedPercentiles`, while the other uses the ones coming from the ETL process.

This should be an easy fix.

Screen Shot 2020-03-24 at 1 39 22 PM

hamilton:
Closes #345

 - use `transformedPercentiles` for Fx probes in the summary table instead of `percentiles`
- use two-pt decimal percentages in tables where appropriate

### DIFF
--- a/src/components/explore/ComparisonSummary.svelte
+++ b/src/components/explore/ComparisonSummary.svelte
@@ -1,5 +1,4 @@
 <script>
-import { format } from 'd3-format';
 import Tweenable from 'udgl/data-graphics/motion/Tweenable.svelte';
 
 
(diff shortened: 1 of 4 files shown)


src/components/explore/ComparisonSummary.svelte
```diff
@@ -2,7 +2,6 @@
 import { format } from 'd3-format';
 import Tweenable from 'udgl/data-graphics/motion/Tweenable.svelte';
 
-let pFmt = format('.0%');
```
openjck, line 5: After this is removed, we get:

`  2:10  error  'format' is defined but never used                               no-unused-vars`

openjck (changes requested):
Looks good to me aside from that!

src/components/explore/ComparisonSummary.svelte
```diff
@@ -2,7 +2,6 @@
 import { format } from 'd3-format';
```
openjck:
```suggestion
```

hamilton:
Thanks for the spot check @openjck ~

PR #356, opened 2020-03-26
Fix #338, fix #355: Fix path auth, browser nav

### LINKED ISSUE #338: Cannot start authentication flow on any page with a path
## Steps to reproduce

1. Open a private browsing window
2. Visit any page with a path, e.g. http://localhost:5000/firefox/probe/gc_ms/explore

## Expected result

Login flow. Provide username and password to Auth0, etc.

## Actual result

> **Oops!, something went wrong**
>
> Callback URL mismatch.
>
> The provided redirect_uri is not in the list of allowed callback URLs.
>
> Please go to the Application Settings page and make sure you are sending a valid callback url from your application.

### LINKED ISSUE #355: Cannot navigate back to probe after clicking logo
## Steps to reproduce

1. Visit any probe, e.g. https://glam-stage.bespoke.nonprod.dataops.mozgcp.net/firefox/probe/gc_ms/explore
2. Click the GLAM logo at the top-left
3. Use the browser's back button

## Expected result

The probe that was previously loaded (in this case, gc_ms) loads again.

## Actual result

The probe that was previously loaded (in this case, gc_ms) does not load.

openjck:


src/routing/Router.svelte
```diff
@@ -1,16 +1,3 @@
...
-  // function to update reactively.
-  export function getPath(path, query) {
-    return query ? `${path}?${query}` : path;
-  }
-
-  export function navigate(...args) {
-    page.show(getPath(...args));
-  }
-</script>
-
```
openjck, line 13: Here's the deal. For this bugfix, I wanted a way to navigate without specifying the path and the query separately. Rather than refactoring `getPath` and `navigate` to reinvent the wheel, I decided that it was easier to get rid of them. Using template literals and `page.show` directly is just as easy.

src/routing/Router.svelte
```diff
@@ -28,8 +15,10 @@
   import NotFound from './pages/NotFound.svelte';
 
 
-  let component;
+  const ssAuthKey = 'relativeURLBeforeAuth';
```
openjck, line 18: 🛳 SS Auth Key \***toot**\* \***toot**\*

openjck: I almost added this as a source comment but that would be unprofessional. Duh.

src/components/controls/ProbeViewControl.svelte
```diff
@@ -32,7 +32,7 @@ export let transformed = options.map((opt) => ({
     multi={false}
     level="medium"
     on:selection={(evt) => {
-      navigate(`/${$store.route.product}/${$store.route.section}/${$store.route.probeName}/${evt.detail.selection}`, $currentQuery);
+      page.show(`/${$store.route.product}/${$store.route.section}/${$store.route.probeName}/${evt.detail.selection}?${$currentQuery}`);
```
spasovski, line 35: What's the difference? Guessing in theory they do the same thing.

src/components/regions/GLAMMark.svelte
```diff
@@ -16,7 +15,7 @@ a {
 }
 </style>
 
-<a href={getPath('/', $currentQuery)} on:click={store.reset}>
+<a href={`/?${$currentQuery}`} on:click={store.reset}>
```
spasovski, line 18: This is a separate point but I think this is a bit of an anti-pattern. I'd argue the logo should always do a `href="/"` with the hard refresh and everything. Anyway it's fine within this PR.

src/routing/Router.svelte
spasovski: Ah this answers my first question.

spasovski (approved):
Looks good at a glance though I don't have much experience with page.js

src/components/regions/GLAMMark.svelte
openjck: Sweet. Please open an issue where we can discuss.

hamilton (approved):
seems to work as expected – thanks @openjck.

PR #360, opened 2020-03-26
fixes #359 - fix posting to probes/random on search

### LINKED ISSUE #359: Search input causes POST to random probes API
Steps to reproduce...

1. Open Glam
2. Open Network monitor
3. Move input to probe search box and each change event triggers a POST to the random probes API

spasovski:


src/routing/pages/Home.svelte
```diff
@@ -12,6 +12,9 @@
 
   // TODO: add this to the upcoming config.js
   const NUMBER_OF_RANDOM_PROBES = 9;
+
+  let selectedProcess;
```
openjck: It looks like this `let` isn't necessary.

openjck:
This is strange. It works, however! Maybe Hamilton could explain why

Other than that one thing, this is r+!

openjck (changes requested):
r+wc

PR #363, opened 2020-03-27
fixes #357 - show pretty errors on 4xx as well

### LINKED ISSUE #357: catch unbecoming 40* error
We seem to be not catching 400 errors in human-readable language. See `checkerboard_severity` for an example. This should be easy to catch and throw accordingly.

I would opt to actually say something like `403: this probe cannot be accessed` or something like that. The below screenshot is from the staging server.

Screen Shot 2020-03-26 at 7 07 54 AM

spasovski:
image


openjck:
👍

PR #366, opened 2020-03-27
Fix #365: Fix probe page loading issue

### LINKED ISSUE #365: Navigating to probe directly causes page not to load
e.g., https://glam-stage.bespoke.nonprod.dataops.mozgcp.net/firefox/probe/gc_ms/explore presents a blank screen

openjck:


openjck:
This the underlying bug a little tricky to reproduce locally. Sometimes loading http://localhost:5000/firefox/probe/gc_ms/explore on master presents a blank screen, sometimes it doesn't. Enabling throttling in the *Network* tab might help.

In either case, we **do not** want to see a blank page when loading that page on this branch.

src/state/store.js
```diff
@@ -277,17 +267,10 @@ export function fetchDataForGLAM(params) {
   );
 }
 
-function intersection(a, b) {
-  const bSet = new Set(b);
-  return new Set(
-    [...a].filter((x) => bSet.has(x)),
-  );
-}
-
```
openjck, line 286: This is unused. It doesn't really have anything to do with this PR, but I was in the area.

src/components/regions/ProbeDetails.svelte
```diff
@@ -22,8 +22,8 @@ const PROBE_TYPE_DOCS = {
 let visible = false;
 onMount(() => { visible = true; });
 
-function probeIsSelected(probe) {
-  return probe.name !== null && probe.name !== 'null';
+function probeIsSelected(p) {
```
spasovski, line 25: the other way should've been fine too but I really like this for human readability between this and the import, also consider `probeItem` or `probeObj`

src/state/store.js
```diff
@@ -111,16 +103,10 @@ const initialState = {
...
-  // get matching probe here
-  const probe = get(probeSet).find((d) => d.name === name);
-  store.setField('probe', probe);
-};
-
 store.reset = () => {
   store.reinitialize({
     auth: store.getState().auth,
-    probe: { name: undefined },
+    probeName: undefined,
```
spasovski, line 109: I like uninitialized strings to be the empty string `''` so it's still set to the same type.

spasovski (approved):
Looks good as-is, the comments are nitpicks.

src/components/regions/ProbeDetails.svelte
openjck: The original way raised an eslint warning. I agree `p` isn't the best name, so I'll update it. Thanks!

src/state/store.js
openjck: Good call! Yes, this would match the above. Thank you!

hamilton (approved):
This looks good to me @openjck, & works as expected

PR #382, opened 2020-04-02
fixes #370 add invalid process selected error

### LINKED ISSUE #370: Show "This probe doesn't exist in this process" when appropriate
If a user ends up looking at a probe while their state is set to a process that the probe doesn't support we should show a page that indicates as such. But we need to be careful to not hard code 'process' so we can support this for non-Fx desktop products against other dimensions.

spasovski:
image

### DIFF
--- a/src/components/regions/MainSelectors.svelte
+++ b/src/components/regions/MainSelectors.svelte
@@ -73,15 +73,17 @@ const COMPACT = true;
         </MenuList>
     </div>
   </MenuButton>
-  <MenuButton tooltip={'Select a Process'} compact={COMPACT}  offset={OFFSET} location='bottom' alignment='left'>
-    <div class=main-filter__label slot="label">{getFieldValueLabel('process', $store.process)}<div class=pull-right-edge><DownCarat size=14 /></div></div>
-    <div slot="menu">
-        <MenuList on:selection={(event) => { store.setField('process', event.detail.key); }}>
-          {#each CONFIG.fields.process.values as {key, label}, i (key)}
-          <MenuListItem  key={key} value={key}><span class='story-label
-            first'></span>{label}</MenuListItem>
-          {/each}
-        </MenuList>
-    </div>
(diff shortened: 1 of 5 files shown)


openjck:
This brings our ESLint problem count down by 2! 🎉

openjck:
I notice that if I use an invalid process (e.g., `process=turtle`), the dropdown defaults to "select a process", which is great, but random probes also do not load on the homepage. Would it make sense to fall back to the first available process in this case? Not sure how we handle this elsewhere.

src/routing/pages/probe/Table.svelte
```diff
@@ -5,21 +5,27 @@
...
+            probeType={probeType}
+            colorMap={data.bucketColorMap}
+            visibleBuckets={[...$store.activeBuckets]}
+            aggregationLevel={$store.aggregationLevel}>
+              <ProbeTitle />
+          </ProbeTableView>
+        {/if}
+      {:else}
+        <DataError reason={`This probe does not record in the ${$store.process} process.`} />
       {/if}
```
openjck: If I'm not mistaken, it looks like lines 15, 27, 28, and 29 can be moved to *Probe.svelte* so that we're not repeating that logic in both *Explore.svelte* and *Table.svelte*.

src/routing/pages/probe/Explore.svelte
```diff
@@ -22,36 +24,40 @@
...
+        <div class="graphic-body__content">
+          <div style="width: 100%">
+            <Spinner size={48} color={'var(--cool-gray-400)'} />
+          </div>
         </div>
-      </div>
+      {/if}
+    {:else}
+      <DataError reason={`This probe does not record in the ${$store.process} process.`} />
     {/if}
```
openjck: Same here. It looks like this can be moved to *Probe.svelte*.

openjck (approved):
Looks good to me! 👍

PR #384, opened 2020-04-03
Tweak how booleans are returned in API responses

robhudson:


hamilton:
@robhudson I pushed a commit to this branch that makes the boolean scalars load like categorical plots.

PR #387, opened 2020-04-06
Allow random API to accept other query params

robhudson:


glam/api/views.py
```diff
@@ -261,14 +261,22 @@ def probes(request):
 @api_view(["POST"])
 def random_probes(request):
     n = request.data.get("n", 3)
-    process = request.data.get("process", "content")
+    channel = request.data.get(
```
spasovski, line 264: good call

spasovski (commented):
looks great thanks

PR #393, opened 2020-04-09
Fixes the case where keyed enumerated histograms throw error when certain buckets are missing from a keyed set

### LINKED ISSUE #217: for every key, make sure the bins are consistent across all aggregations after the data is fetched
We have cases where keys w/ sparse data might not necessarily share the same buckets with their siblings. One such case is `network_cache_entry_count_share` – in my demo data, the `WASM` key has fewer buckets than. the rest, which breaks the UI since we assume otherwise.

hamilton:
Fixes #217. Turns out this was error caused by an old data workaround that is no longer valid.

### DIFF
--- a/src/components/explore/ProbeExplorer.svelte
+++ b/src/components/explore/ProbeExplorer.svelte
@@ -294,7 +294,7 @@ $: if (hoverValue.x) {
     />
   </div>
   <div style="display: {insufficientData ? 'none' : 'block'}">
-    <CompareClientVolumeGraph 
+    <CompareClientVolumeGraph
       data={clientCountsData}
       description={compareDescription(clientVolumeOverTimeTitle)}
       yDomain={yClientsDomain}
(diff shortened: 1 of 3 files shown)


openjck:
This brings our ESLint problem count down by 2, from 199 to 197! :tada:

openjck:
I'm seeing *"Oh no! Your client sent an invalid request."* on the following page, even on this branch. Is that expected?

http://localhost:5000/firefox/probe/popup_notification_stats/explore?activeBuckets=%5B%5D&aggregationLevel=build_id&channel=nightly&os=Windows&process=content&proportionMetricType=proportions&timeHorizon=MONTH&versions=%5B%5D&visiblePercentiles=%5B95%2C75%2C50%2C25%2C5%5D

openjck:
Looks good to me. No console errors. r+ 🎉

openjck (approved):

PR #404, opened 2020-04-10
explicitly deal with probe fetch 404s

spasovski:
Wording of this is up for debate.

image

### DIFF
--- a/src/state/api.js
+++ b/src/state/api.js
@@ -5,6 +5,7 @@ const randomProbeURL = '__BASE_DOMAIN__/api/v1/probes/random/';
 const FETCH_ERROR_MESSAGES = {
   code5xx: 'Oh no! The server encountered an error.',
   code4xx: 'Oh no! Your client sent an invalid request.',
+  code404: 'No data found for the selected dimensions.',
 };
 
 export async function getRandomProbes(numProbes, process) {
@@ -29,7 +30,12 @@ export async function getProbeData(params, token) {
   }).then(async (response) => {
     // catch fetch error responses and surface them into
     // the frontend.
-    if (response.status >= 400 && response.status < 600) {
+    if (response.status === 404) {
+      const e = new Error(FETCH_ERROR_MESSAGES.code404);
+      e.moreInformation =
+        "You may try changing the probe's current data filters.";
+      throw e;
+    } else if (response.status >= 400 && response.status < 600) {
       const msg =
         response.status < 500
           ? FETCH_ERROR_MESSAGES.code4xx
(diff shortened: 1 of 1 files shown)


openjck:
#404 is about 404s? 🙃

src/state/api.js
```diff
@@ -5,6 +5,7 @@ const randomProbeURL = '__BASE_DOMAIN__/api/v1/probes/random/';
 const FETCH_ERROR_MESSAGES = {
   code5xx: 'Oh no! The server encountered an error.',
   code4xx: 'Oh no! Your client sent an invalid request.',
+  code404: 'No data found for the selected dimensions.'
```
openjck: There's a new Prettier error on this line. It looks like it wants a trailing comma. (Trailing commas are a new default in Prettier 2.0.0.)

src/state/api.js
```diff
@@ -29,7 +30,11 @@ export async function getProbeData(params, token) {
   }).then(async (response) => {
     // catch fetch error responses and surface them into
     // the frontend.
-    if (response.status >= 400 && response.status < 600) {
+    if (response.status === 404) {
+      const e = new Error(FETCH_ERROR_MESSAGES.code404);
+      e.moreInformation = "You may try changing the probe's current data filters.";
```
openjck: There's a Prettier error on this line. It looks like it wants the line to be shorter.

In both cases, simply running `npm run format` should do the formatting for you.

src/state/api.js
```diff
@@ -39,7 +44,7 @@ export async function getProbeData(params, token) {
       e.moreInformation = `GLAM encountered a problem trying to show ${
         params.probe
       }, so we'll
-      report the error here: 
+      report the error here:
```
openjck, line 48: EditorConfig! 🎉

openjck (changes requested):
This looks good to me, aside the Prettier things. Pretty straightforward and the wording is fine with me!

src/state/api.js
spasovski: How do you run the linter?

src/state/api.js
spasovski: heh

src/state/api.js
openjck: You can run `npm run format` from the command line. Another option is to configure your editor to auto-run Prettier on all GLAM JS files. That's what I've set up in my editor.

openjck: I wouldn't configure your editor to auto-run Prettier on Svelte files yet because it's got some bad known issues, including one that causes it to return garbage when Prettier 2.0.0+ is installed.

openjck: https://github.com/UnwrittenFun/prettier-plugin-svelte/issues/84

PR #406, opened 2020-04-14
add new config-js structure

spasovski:
I had something more complicated locally that also moves much of the helper methods in store.js like `getField() getFieldValues() isField()` but decided against it until for now until we're certain of the direction.

### DIFF
--- a/src/config/firefox-desktop.js
+++ b/src/config/firefox-desktop.js
@@ -1,0 +1,41 @@
+export default {
+  sampleRate: .1,
+  dimensions: {
+    os: {
+      title: 'Operating System',
+      key: 'os',
+      values: [
+        { key: 'ALL', label: 'All OSes', keyTransform: 'NULL' },
+        { key: 'Windows', label: 'Windows' },
+        { key: 'Mac', label: 'Mac' },
+        { key: 'Linux', label: 'Linux' },
+      ],
+    },
+    channel: {
+      title: 'Channel',
+      key: 'channel',
+      values: [
+        { key: 'nightly', label: 'Nightly' },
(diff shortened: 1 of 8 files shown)


src/config/firefox-desktop.js
```diff
@@ -0,0 +1,53 @@
...
+  },
+  versions: {
+    title: 'Versions',
+    key: 'version',
+    type: 'multi',
+    skipValidation: true,
+    values: [
+      { key: 'nightly', label: 'Nightly' },
+      { key: 'beta', label: 'Beta' },
+      { key: 'release', label: 'release' },
```
robhudson: These aren't versions.

src/state/store.js
```diff
@@ -160,31 +162,38 @@ export const hasDefaultControlFields = derived(store, ($store) =>
...
+    const params = getParamsForQueryString(obj);
+    delete params.timeHorizon;
+    delete params.proportionMetricType;
+    delete params.activeBuckets;
+    delete params.visiblePercentiles;
+    params.probe = obj.probeName;
+    params.os = osValue;
+    params.channel = channelValue;
+    params.process = process;
+    params.versions = []; // TODO: This should not be expected by the API.
```
spasovski: @robhudson we found that this param is still expected by the API, probably something to flag for future removal

hamilton: @spasovski let's file an issue for this so we can track that separately.

hamilton (approved):
Generally looks good. We're not quite ready to generalize for Fenix yet but this is a great step in the right direction. Thanks @spasovski.

PR #407, opened 2020-04-14
add remaining color palettes

spasovski:
Palettes had their saturation increased by 5% for values < 500 (except cool-gray/pond-green) and were converted back to HSL for easier numerical adjustment going forward. There will be visual differences but hopefully not deal-breakers.

Left is old, right is new:
image

### DIFF
--- a/public/static/global.css
+++ b/public/static/global.css
@@ -168,14 +168,14 @@
     --body-gray-02: var(--cool-gray-900);
 
     /* use digital blue as link + button color for accessibility */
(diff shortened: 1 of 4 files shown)


hamilton (commented):
Looks good. I left some comments about saturating the digital blues in slack. Also, I think we can up the 5th percentile color in `color-map.js`, but otherwise everything checks out. Feel free to work through this bit of feedback and merge at your discretion.

PR #415, opened 2020-04-17
fixes #412 make fetch error messages simpler

### LINKED ISSUE #412: make fetch error messages more detailed
Currently 405s don't have a response body so nothing gets shown. In this case we could show our own error string along with the status code.

spasovski:
image

### DIFF
--- a/src/state/api.js
+++ b/src/state/api.js
@@ -5,7 +5,8 @@ const randomProbeURL = '__BASE_DOMAIN__/api/v1/probes/random/';
 const FETCH_ERROR_MESSAGES = {
   code5xx: 'Oh no! The server encountered an error.',
   code4xx: 'Oh no! Your client sent an invalid request.',
-  code404: 'No data found for the selected dimensions.',
+  code404: '404: No data found for the selected dimensions.',
+  code405: '405: Method not allowed.',
 };
 
 export async function getRandomProbes(numProbes, process) {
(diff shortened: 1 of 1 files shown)


src/state/api.js
```diff
@@ -28,26 +29,17 @@ export async function getProbeData(params, token) {
...
-      const txt = await response.text();
-      e.moreInformation = `GLAM encountered a problem trying to show ${
-        params.probe
-      }, so we'll
-      report the error here:
-      ${txt.split('\n').slice(0, 2)}`;
-      throw e;
+    // Catch fetch error responses and show them in the UI.
+    if (response.status >= 400 && response.status < 600) {
+      const errorCode = 'code' + response.status;
```
openjck:
```
  34:25  error  Unexpected string concatenation  prefer-template
```

spasovski: That's a little strict, simple concatenations shouldn't need it. Don't mind changing it just weird.

openjck: Yeah, Airbnb rules. 🤷 

We could change that rule if we prefer.

openjck (changes requested):
Looks good to me other than the eslint thing. 👍

src/state/api.js
hamilton: I do prefer string templating in cases like this

PR #418, opened 2020-04-18
Move dimension filters to body header; automatically select a process if current one is invalid; design tweaks to accomodate these changes

### LINKED ISSUE #400: move dimension menu items to be in body header, not app header
Screen Shot 2020-04-09 at 1 39 15 PM

### LINKED ISSUE #401: need dimension menu boxes to be dynamically read from config files

hamilton:
Closes #400, closes #401


src/routing/wrappers/Layout.svelte
```diff
@@ -20,11 +20,13 @@
   <Toolbar sticky>
     <GLAMMark />
     <Search />
-    <MainSelectors />
+    <div style="justify-self: end; padding-right: var(--space-4x); min-width: var(--space-16x);">
```
hamilton, line 23: this is a placeholder until we add a product selector.

src/components/regions/MainSelectors.svelte
```diff
@@ -37,53 +40,43 @@ const COMPACT = true;
...
-</div>
+{#if $store.route.section === 'probe' && $probe}
+  <div transition:fly={{ x: 5, duration: 200 }} class='main-filters'>
+    {#each Object.values(productConfig[$store.product].dimensions) as dimension, i (dimension.key)}
+      {#if dimension.values.some(
+            (di) => dimension.isValidKey === undefined
+                    || dimension.isValidKey(di.key, $probe, store),
+      )}
+      <MenuButton tooltip='Select a {dimension.title}' compact={COMPACT} offset={OFFSET} location='bottom' alignment='right'>
+        <div class=main-filter__label slot="label">
```
spasovski, line 66: `"main-filter-label"` might be better if we plan to stick with things like `"main-filter-label--active"` etc going forward otherwise we'd have to memorize when to use `_` vs `-`

spasovski: I think it was already like this so it shouldn't hold up this PR. I can do a cleanup PR for things like this later.

src/config/firefox-desktop.js
```diff
@@ -24,18 +26,34 @@ export default {
...
+  setDefaultsForProbe(store, probe) {
+    const state = store.getState();
+    if (
+      !isSelectedProcessValid(
+        probe.record_in_processes,
+        state.productDimensions.process
+      )
+    ) {
+      let newProcess = probe.record_in_processes[0];
+      if (newProcess === 'main') newProcess = 'parent';
```
spasovski, line 55: still sucks that we have to do this

src/config/firefox-desktop.js
```diff
@@ -24,18 +26,34 @@ export default {
...
   },
+  setDefaultsForProbe(store, probe) {
+    const state = store.getState();
+    if (
+      !isSelectedProcessValid(
+        probe.record_in_processes,
+        state.productDimensions.process
+      )
+    ) {
+      let newProcess = probe.record_in_processes[0];
```
spasovski, line 54: Is it OK to always set it to the first item from the API? Perhaps we should default to parent.

src/config/products.js
```diff
@@ -0,0 +1,5 @@
+import FIREFOX_DESKTOP from './firefox-desktop';
+
+export default {
+  firefoxDesktop: FIREFOX_DESKTOP,
```
spasovski, line 4: Is the idea that the product dropdown will have `Object.keys(PRODUCTS)`?

src/udgl/menu/MenuButton.svelte
```diff
@@ -35,18 +35,26 @@ let button;
...
-  color: white;
+  border: 1px solid var(--cool-gray-200);
+  color: var(--digital-blue-700);
+  background-color: var(--cool-gray-subtle);
+  box-shadow: 0px 2px 2px rgba(0,0,0,.05);
+  transition: 100ms;
+  cursor: pointer;
+}
+
+.activating-button:hover {
```
spasovski, line 47: Will these interfere with the GP Button work by @openjck ?

src/config/firefox-desktop.js
hamilton: Not sure. From inspection, I found that `parent` is always the first if `parent` is available to the probe.

src/udgl/menu/MenuButton.svelte
hamilton: This is a stopgap since his PR has not been revised yet. I didn't want to block on it.

hamilton: This all said, I'll move this `MenuButton` component out of `udgl` and into `components` for now.

src/components/regions/MainSelectors.svelte
hamilton: Since we're roughly following hte BEM methodology, I think it should be __ for a conceptual child and -- for a variant.

src/config/products.js
hamilton: Yes, & that suggests some additions to each per-product config file once we get to that

openjck:
This brings our ESLint count down from 113 to 107. 🎉

src/udgl/menu/MenuButton.svelte
openjck: Don't worry about me; I don't mind rebasing.

src/components/regions/ProbeDetails.svelte
```diff
@@ -243,6 +244,9 @@ h2 {
                 </div>
             {/await}
         </div>
+        <div class='drawer-section  drawer-section--end'>
+          <Button level=low><Doc size={16} /> Documentation</Button>
```
spasovski: This currently does nothing.

spasovski:
I think this brings a lot to the table. The new menu buttons have issues (can't immediately click into a new menu, cursor switches to default between clicks) but less issues than the old ones so not a blocker.

I'd just either remove the doc link for now (or comment it out) until it does something...though I am not sure if the documentation is quite ready yet. r+wc

PR #421, opened 2020-04-20
moves from store.aggregationLevel to store.productDimensions.aggregationLevel in Explore.svelte

### LINKED ISSUE #413: the buttons in the Probe body throws an error and renders weirdly when clicked

hamilton:
Fixes #413

### DIFF
--- a/src/routing/pages/probe/Explore.svelte
+++ b/src/routing/pages/probe/Explore.svelte
@@ -8,15 +8,17 @@
   import ProbeTitle from '../../../components/regions/ProbeTitle.svelte';
   import { store, probe } from '../../../state/store';
 
-
   function handleBodySelectors(event) {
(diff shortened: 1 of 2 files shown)


src/routing/pages/probe/Explore.svelte
```diff
@@ -43,7 +45,7 @@
         timeHorizon={$store.timeHorizon}
         percentiles={$store.visiblePercentiles}
         on:selection={handleBodySelectors}
-        aggregationLevel={$store.aggregationLevel}>
+        aggregationLevel={$store.productDimensions.aggregationLevel}>
```
spasovski, line 48: Good catch. This needs to happen in `Table.svelte` as well though.

spasovski (changes requested):
r+wc

PR #422, opened 2020-04-20
fixes #419 get aggregationLevel properly

### LINKED ISSUE #419: "by major version" is returning a duplicate build id error in the frontend
The API looks to be returning data successfully but the front-end says "Duplicate Build IDs found" which, technically, is true -- they are all `"*"` from the API, but that's intended when choosing this menu item.

spasovski:
fixes #419


hamilton:
Awaiting review / merge of #421 , which blocks review here

hamilton (approved):

PR #429, opened 2020-04-21
Fix #428: Don't reset store values after auth

### LINKED ISSUE #428: os, process, and other store values are overwritten after authentication
## Steps to reproduce

1. Visit [this page](http://localhost:5000/firefox/probe/gc_ms/explore?activeBuckets=%5B%5D&aggregationLevel=build_id&channel=nightly&os=Windows&process=content&proportionMetricType=proportions&timeHorizon=MONTH&visiblePercentiles=%5B95%2C75%2C50%2C25%2C5%5D) or any other probe page
2. Set the OS to *Mac* and the process to *Parent*
3. Copy the URL
4. Paste that URL into private browsing mode
5. Log in

## Actual result

OS is back to its default (*Windows*) and process is back to its default (*Content*)

## Expected result

OS remains *Mac* and process remains *Parent*

openjck:
Fixes #428

### DIFF
--- a/src/routing/Router.svelte
+++ b/src/routing/Router.svelte
@@ -3,9 +3,11 @@
   import Spinner from 'udgl/LineSegSpinner.svelte';
   import { onMount } from 'svelte';
 
+  import productConfig from '../config/products';
   import { store, currentQuery } from '../state/store';
   import { probeSet } from '../state/telemetry-search';
-  import productConfig from '../config/products';
+  import { codeAndStateInQuery } from '../utils/url';
+
   // Wrappers
   import Layout from './wrappers/Layout.svelte';
 
@@ -16,9 +18,7 @@
   import NotFound from './pages/NotFound.svelte';
 
 
-  const ssAuthKey = 'relativeURLBeforeAuth';
   let visible = false;
-  let handledAuthRedirect = false;
   let component;
 
   function updateQueryString(query) {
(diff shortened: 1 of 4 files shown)


openjck:
For whatever its worth, this also makes auth a lot simpler. Win-win!

src/routing/Router.svelte
```diff
@@ -29,42 +27,22 @@
...
-  // speaking, we don't need to check $store.auth.isAuthenticated here because
-  // it should always be true when handledAuthRedirect is true, but it doesn't hurt
-  // to double check.)
-  $: if (handledAuthRedirect && $store.auth.isAuthenticated && visible) {
+  // We also need to avoid running this block whenever the code= and state= keys
+  // are in the query, otherwise we would clobber them. Auth0 needs to read them
+  // to authenticate the user.
+  $: if (
+    visible
+    && !window.location.search.includes('code=')
```
spasovski: this feels brittle but i see it was always present

src/utils/auth.js
```diff
@@ -22,51 +22,55 @@ async function withToken(cb) {
...
+  window.location.replace(realAuthRedirectTarget);
+}
+
 export function authenticate(successCallback) {
   window.onload = async () => {
     await configureClient();
 
     const isAuthenticated = await auth0.isAuthenticated();
+    const codeAndStatePresent =
+      window.location.search.includes('code=') &&
```
spasovski: good candidate for a util function, we repeat this in 2 places

src/routing/Router.svelte
openjck: How do you mean? I'm happy to improve it. 😃

src/utils/auth.js
openjck: Good call!

openjck:
Updated! Let me know what you mean about brittle. Happy to fix.

openjck:
Davor also mentioned on Slack that visiting the homepage in a private browsing tab brings up an error about an operation being aborted.

openjck:
I'm really not sure what to make of the `AbortError`. There's not much documentation about it online, other than some reports that it's a bug in Firefox that sometimes manifests when scrubbing videos. It could have something to do with the fact that the chart loading on the homepage is being interrupted.

I don't see anything like that error in Chrome.

Given that it doesn't seem to cause any material problems, it doesn't happen in Chrome, and it may (???) be related to a bug in Firefox, I think we should merge this to fix auth. But I'm open to getting a second opinion if you prefer.

spasovski:
Looks great thanks! The abortError is indeed odd and doesn't actually happen 100% of the time - I vote we get this in.

PR #439, opened 2020-04-23
Allow bucket to be passed as an argument for import_aggs.

emtwo:
This will help us test new data locally before merging changes in the ETL. I would place data in a project and bucket called `glam-desktop-dev`. So we can pull in data like this:

```
export GOOGLE_CLOUD_PROJECT=glam-desktop-dev
./manage.py import_aggs release --bucket glam-desktop-dev
```


robhudson:
Looks good. Can you add a small note to the docs that this is an option?

PR #462, opened 2020-05-04
Update database to have a column for percentiles and aggregations

robhudson:
This requires the extract process to join 2 rows into a single row with 2 columns. The advantage is that we can query all the data we need given a set of dimensions and get a single row.

Flagging this as a draft pull request as this will require coordination with the bigquery-etl and telemetry-airflow repos for the upstream changes as well.


robhudson:
Closing in favor of the transpose-schema branch.

PR #471, opened 2020-05-05
basic search functionality from probe search service

### LINKED ISSUE #172: move search out of client and into a probe search service
When we started building GLAM, it was unclear what the server architecture would be. So we opted to do all the search client side, which is limited for the volume of probes we're indexing. Since we've moved to django, perhaps a postgres search API will be a huge improvement.

Ideally, however, we can solve this issue for a handful of data tools at once. The probe dictionary would benefit from this as well. The right solution might actually look more like a search API we can consume from both GLAM and the probe dictionary.

hamilton:
closes #172

@robhudson this is a first pass and not ready for review. There are issues with (1) casing, and (2) the data API doesn't accept all the same strings as the probe search service gives. We should work to unify these. Preferably GLAM would match much more closely what is in the probe search service, since it seems people have been tripped up by GLAM's naming convention during the interviews.

Work left on this PR:

- [ ] remove all the bits of global state we were tracking before. This seems unnecessary in hindsight since the search workflow is 100% confined to the component. This will remove some complexity and improve maintainability.
- [ ] fix the keyboard interactions
- [ ] make the search experience somewhat nicer
- [ ] figure out the solution to the name mismatch b/t the probe search service & GLAM

### DIFF
--- a/src/components/search/Search.svelte
+++ b/src/components/search/Search.svelte
@@ -16,6 +16,7 @@ import TelemetrySearchResults from './SearchResults.svelte';
 
 let inputElement;
 let searchContainer;
+let results = [];
 
 function turnOnSearch() {
     store.setField('searchIsActive', true);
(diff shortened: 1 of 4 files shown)


openjck:
Superseded by #611

PR #472, opened 2020-05-05
fixes #463 move details drawer to the left 

### LINKED ISSUE #463: move the probe details column to the left of the content column
Stretch goal...have it tuck away below the content on smaller viewports.

spasovski:
image

### DIFF
--- a/public/static/global.css
+++ b/public/static/global.css
@@ -570,10 +570,12 @@ header h1 div span {
   width: var(--drawer-width);
 }
 
-.right-drawer {
+.details-drawer {
   background-color: var(--cool-gray-subtle);
-  height: 100%;
-  border-radius: 0 var(--content-border-radius) var(--content-border-radius) 0;
+  height: calc(100% - var(--space-3x));
+  border-radius: var(--content-border-radius) 0 0 var(--content-border-radius);
+  padding: var(--space-3x) var(--space-2x) 0 var(--space-2x);
+  box-sizing: border-box;
 }
 
 .left-drawer__header {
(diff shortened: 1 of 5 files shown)


public/static/global.css
```diff
@@ -891,10 +893,6 @@ tr .data-cell--header:first-child {
   text-align: center;
 }
 
-/* default */
-.data-cell--align-undefined {
-}
```
openjck, line 896: 🎉

src/routing/wrappers/Layout.svelte
```diff
@@ -29,9 +32,9 @@
       <MainSelectors />
     </ContentHeader>
     <ContentBody>
-      <div class="graphic-body">
+      <div class="graphic-body" class:graphic-body-home={isHomePageView}>
         <slot />
-        {#if $store.route.section === 'probe'}
+        {#if !isHomePageView}
```
openjck: What's the benefit of reversing this logic? We're rendering something that only makes sense in the probe section.

src/routing/wrappers/Layout.svelte
```diff
@@ -14,6 +14,9 @@
   import ProbeDetails from '../../components/regions/ProbeDetails.svelte';
 
   import { store } from '../../state/store';
+
+  let isHomePageView = true;
```
openjck: It looks like we don't need this `let`.

openjck (changes requested):
r+wc

src/routing/wrappers/Layout.svelte
spasovski: Yea I suppose this should be `isProbeDetailsView` instead so we don't have to check the inverse.

PR #473, opened 2020-05-05
make homepage documentation link active

spasovski:


src/components/home/MarketingBlock.svelte
```diff
@@ -83,11 +83,13 @@ $: $t = 1;
     </div>
     <!-- <img src='bg.jpg' alt='yes' /> -->
     <div class='marketing-block__call-to-action'>
-      <Button tooltip="coming soon!">Documentation</Button>
+      <a target="_blank" href="https://docs.google.com/document/d/1qIkEDemnODbYuVIfpciohgEXcaFjrK_mfVG5FOSeVuM/edit?usp=sharing">
```
openjck: Should we link to the *Viewing* mode instead? When I load this URL, I see a bunch of comments in the sidebar. I'm not sure if that's useful and it may even be a bad look.

openjck: There's also a subtle underline below the word *Documentation* when the button is hovered.

openjck (changes requested):
r+wc

PR #475, opened 2020-05-06
fixes #466 send all-oses param properly

### LINKED ISSUE #466: there is no data for the "all OSes" channel
The "All OSes" functionality is, well, non-functional; I haven't encountered a single probe that has any data for this dimension value. We should remove it and investigate.

spasovski:
fixes #466


hamilton (approved):
lgtm – this is a nice reduction in code and seems to work locally. Thanks @spasovski

PR #477, opened 2020-05-07
Include "prettier --check" when running "npm test"

openjck:


.circleci/config.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           paths:
             - ./node_modules
       - run: npm run build
-      # - run: npm run test # don't run this until design library is gone
+      # - run: npm test # don't run this until design library is gone
```
openjck, line 46: Not strictly related to this PR (the code is commented after all, so it's not doing much), but I happened to notice this. We don't need `run` to run tests.

PR #478, opened 2020-05-07
fixes #467 add data error alert notice

### LINKED ISSUE #467: add some kind of cancelable notice on the bottom left saying that staging and dev will still have bugs / data validation issues

spasovski:
Do `localStorage.clear()` in your console to see again after dismissing.

- Not super happy with the text copy and possibly color scheme so feel free to suggest better.

image

Updated with John's copy:

image


src/components/controls/AlertNotice.svelte
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  .alert-notice-action button:hover {
+    background-color: var(--bright-yellow-200);
+    color: var(--bright-yellow-750);
+  }
+</style>
+
+{#if alertVisible}
+  <div class="alert-notice" transition:fade={{easing: cubicOut}}>
```
hamilton: I might suggest `transition:fly` with a `y: 10` or so and a `duration` of maybe 200?

openjck:
```
  65:46  error  A space is required after '{'   object-curly-spacing
  65:63  error  A space is required before '}'  object-curly-spacing
```

src/components/controls/AlertNotice.svelte
```diff
@@ -0,0 +1,73 @@
...
+  import { cubicOut } from 'svelte/easing';
+
+  // So we don't pollute the entire LS namespace.
+  const KEY_PREFIX = 'alertNotice-';
+  export let dismissText = 'dismiss'; // Label of 'dismiss' button.
+
+  // Required: Which localStorage key to check whether this notice has been dismissed.
+  export let toggleKey = '';
+
+  $: alertVisible = window.localStorage.getItem(KEY_PREFIX + toggleKey) !== 'true';
```
openjck: In Slack, Rob Miller mentioned that he'd like the message to reappear daily if it's not too much engineering effort, or if that would be too much engineering effort, to use `sessionStorage` instead.

src/components/controls/AlertNotice.svelte
```diff
@@ -0,0 +1,73 @@
+<script>
+  import { fade } from 'svelte/transition';
+  import { cubicOut } from 'svelte/easing';
+
+  // So we don't pollute the entire LS namespace.
+  const KEY_PREFIX = 'alertNotice-';
+  export let dismissText = 'dismiss'; // Label of 'dismiss' button.
```
openjck, line 6: For the sake of consistency, I'd suggest putting all of the props just below the imports.

src/components/controls/AlertNotice.svelte
```diff
@@ -0,0 +1,73 @@
...
+    left: var(--space-4x);
+    bottom: var(--space-4x);
+    border: 1px solid var(--bright-yellow-500);
+    display: flex;
+    background-color: var(--bright-yellow-100);
+    color: var(--bright-yellow-700);
+    border-radius: var(--border-radius-base);
+    padding: var(--space-2x);
+    align-items: center;
+    grid-gap: var(--space-3x);
```
openjck: This doesn't do anything in `display: flex`, does it?

src/components/controls/AlertNotice.svelte
```diff
@@ -0,0 +1,73 @@
...
+    width: fit-content;
```
openjck: Huh. I've never used this!

src/routing/wrappers/Layout.svelte
```diff
@@ -71,4 +72,7 @@
       <Footer />
     </ContentFooter>
   </Content>
+  <AlertNotice toggleKey="dataErrorsWarning">
+    <p>The staging and development instances of GLAM are likely to have data issues and UI bugs.</p>
```
openjck: I actually think this copy is great. The only thing I'd suggest is maybe being more specific about "data issues" if we can.

openjck: I suppose if someone is on dev, they may not care about stage, and vice versa. Maybe something like this.

"This dashboard is under active development. Bugs and [data accuracy issues] may exist."

But I'd defer to Rob for the final copy.

src/components/controls/AlertNotice.svelte
openjck: And yay fade!

src/components/controls/AlertNotice.svelte
```diff
@@ -0,0 +1,73 @@
...
+    font-size: var(--text-015);
+    width: 350px;
+  }
+
+  .alert-notice-content > * {
+    margin: 0;
+  }
+
+  .alert-notice-action button {
+    padding: var(--space-base);
```
openjck: jarek-kalwa-space-base

src/components/controls/AlertNotice.svelte
openjck: What does the `$:` reactivity do in this case? It doesn't seem that manually altering `localStorage` has any immediate effect.

src/components/controls/AlertNotice.svelte
```diff
@@ -0,0 +1,73 @@
...
+
+  function dismissNotice() {
+    alertVisible = false;
+    window.localStorage.setItem(KEY_PREFIX + toggleKey, 'true');
+  }
+</script>
+
+<style>
+  .alert-notice {
+    position: absolute;
```
openjck: Maybe `position: sticky` instead, so that it remains at the bottom-left when the page scrolls?

src/components/controls/AlertNotice.svelte
openjck: Can this be used here? Firefox says it's an invalid value.

src/components/controls/AlertNotice.svelte
```diff
@@ -0,0 +1,73 @@
...
+    left: var(--space-4x);
```
openjck: For some reason, it feels more natural to me on the right. I want to say it's because of some common UI convention or something, but it may just be personal preference.

openjck (commented):
Added some comments

src/components/controls/AlertNotice.svelte
openjck: Oh I guess it does. Guess I need to brush up.

src/components/controls/AlertNotice.svelte
spasovski: Good catch, just an artifact of some earlier testing.

src/components/controls/AlertNotice.svelte
spasovski: Seems a little intrusive but I'll see what the consensus is. Easy to change.

src/components/controls/AlertNotice.svelte
spasovski: Yea another remnant of an earlier approach. Good catch.

src/components/controls/AlertNotice.svelte
spasovski: We're way ahead of Elon!

src/components/controls/AlertNotice.svelte
spasovski: Just a remnant of an earlier approach, and yep I added the check against `today`

src/routing/wrappers/Layout.svelte
```diff
@@ -71,4 +72,7 @@
       <Footer />
     </ContentFooter>
   </Content>
+  <AlertNotice toggleKey="dataErrorsWarning">
+    <p>This dashboard is under active development. Bugs and data accuracy issues may exist.</p>
```
hamilton:
```suggestion
    <p>Thank you for testing the GLAM prototype! 
This tool is still in active development. UX bugs and data issues may exist.
Help us make it suit your needs by directing questions or feedback to the #glam Slack channel.</p>
```

src/components/controls/AlertNotice.svelte
hamilton: we really should make this `--space-1x` across the board.

hamilton (changes requested):
Looking great @spasovski. Thanks for getting this together so quickly.

- I'd love to see a dismiss icon as well. [If only a new icon set could be easily and effortlessly applied, gosh.](https://github.com/graph-paper-org/graph-paper/tree/master/packages/icons) (maybe the `Close` icon). These icons don't require any other stylesheet so they can be used without postcss, so it should be as easy as npm installing the package and then importing the one you want to use. I think the story shows all the icons + the names if you want to peruse.
- I'd like to see a slight shadow to the element, if possible, to bring it a bit off the page. I think we might have some barely-used `box-shadow` variables that could be applied.
- made a suggestion for the copy that gives more context.

src/components/controls/AlertNotice.svelte
hamilton: (I appreciate the instinct to apply a transition!)

Issue #259, opened 2020-02-27
Run Prettier on all Svelte files


openjck:
I haven't had much luck with Svelte plugins for Prettier, but we can try again. I think the code would be a bit easier to read if it were formatted more consistently.

We probably want to land #258 first.

openjck:
prettier-plugin-svelte was recently taken over by the Svelte team and updated to support Prettier 2.x.x. I haven't used it yet and expect that it's still a bit buggy, but it's likely to improve soon.

openjck:
To achieve this, I recommend reproducing [these changes](https://github.com/mozilla/glam/pull/610/commits/c9f8076d7c48f80241780ce49ebeace850baf4e7) against the latest main. Then just run `npm run format`. Simple!

I recommend splitting them into two commits unless you plan to merge immediately, because the effects of `npm run format` will undoubtedly conflict with something before long.

PR #480, opened 2020-05-07
Issue #259: Run Prettier on Svelte files

openjck:


src/components/FirefoxReleaseVersionMarkers.svelte
```diff
@@ -1,33 +1,29 @@
...
       <Marker location={date}>
         {#if labels}
           {#if $store.productDimensions.channel === 'nightly'}
             {label + 2}
           {:else if $store.productDimensions.channel === 'beta'}
             {label + 1}`
-          {:else}
-            {label}
-          {/if}
+          {:else}{label}{/if}
```
spasovski, line 26: This rule is weird. I definitely prefer it the way it was.

openjck: Yeah that is weird. Looking at the prettier-plugin-svelte docs, I don't know if this can be changed. I bet it'll be fixed in a future release, though.

spasovski: It still fixes more than it makes things weird. r+ if everything seems to work locally

openjck:
Superseded by #610

PR #481, opened 2020-05-08
fixes the issue where process=all should default to process=[parent, content, gpu]

### LINKED ISSUE #417: handle firefox desktop probes w/ the "all" process
There are a few cases where we have `all` listed in `probe.record_in_processes`.

The work to close this issue is (1) find out what "all" means in this contexts, and (2) do something either on the probe API side or the client side to transform "all" to something more useful (like listing the processes that are used). I would opt for this to be fixed on the server end, so the data is cleaner going into the frontend.

Here is the list:

```
keypress_present_latency
telemetry_event_recording_error
uptake_remote_content_result_1
browser_usage_graphite
telemetry_event_counts
telemetry_profile_directory_scans
telemetry_accumulate_unknown_histogram_keys
telemetry_keyed_scalars_unknown_keys
```

hamilton:
Closes #417 

We have some high-value probes that mysteriously have `process=all` such as `keypress_present_latency`. This fixes the case by creating a new `transformProbeForGLAM` function within each config file, which does what you would expect. I am imagining that we'll likely encounter issues that require this kind of function on a per-product basis.

### DIFF
--- a/src/config/firefox-desktop.js
+++ b/src/config/firefox-desktop.js
@@ -43,6 +43,13 @@ export default {
       },
     },
   },
+  transformProbeForGLAM(probe) {
+    const pr = { ...probe };
+    if (pr.record_in_processes[0] === 'all') {
+      pr.record_in_processes = ['main', 'content', 'gpu'];
+    }
+    return pr;
+  },
   setDefaultsForProbe(store, probe) {
     const state = store.getState();
     if (
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,14 @@
 ## [unreleased](https://github.com/mozilla/glam/compare/2020.4.2...HEAD) (date TBD)
 
 - adds prototype app warning dialog ([#478](https://github.com/mozilla/glam/pull/478))
+- Fix bug where process=all should convert to process=[parent,content,gpu](<[#481](https://github.com/mozilla/glam/pull/451)>)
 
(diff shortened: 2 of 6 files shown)


src/config/firefox-desktop.js
```diff
@@ -43,6 +43,13 @@ export default {
       },
     },
   },
+  transformProbeForGLAM(probe) {
```
spasovski, line 46: This is a good solution but I still wish it wasn't necessary.

spasovski (approved):
Looks good thanks!

src/config/firefox-desktop.js
robhudson: Perhaps that's a change that could be pushed to probe-scraper, to ultimately end up in probe-info?

hamilton: I think that would be ideal. Do you know how to go about requesting that @robhudson? Is the probe scraper versioned?

robhudson: By the way, there's also `"all_childs"`: https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/collection/histograms.html#record-in-processes

robhudson: `"all_childs"` -> `['content', 'gpu']` (all except `'main'`)

robhudson: https://github.com/mozilla/probe-scraper/issues/187

PR #492, opened 2020-05-15
default to nightly if channel === release && probe only available in prerelease

### LINKED ISSUE #434: default to a valid channel if an invalid channel is selected
For Firefox Desktop - similar to process - probes sometimes don't exist on certain platforms. We need to make sure a relevant channel is selected so data (in theory) loads.

hamilton:
closes #434

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [unreleased](https://github.com/mozilla/glam/compare/2020.5.0...HEAD) (date TBD)
 
+- Select prerelease channel if current channel = release & probe only exists in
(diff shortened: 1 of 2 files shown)


openjck:
What would be a good probe to test this against and/or how might I find a probe like this on my own in the future? What should I be looking for?

hamilton:
@openjck any prerelease probe should suffice, I think. To test, select a probe that _isn't_ prerelease, select `release` for it, then search for a pre-release probe. It should default to `nightly`. on master, it should return thee 'no data' error. https://probes.telemetry.mozilla.org/

openjck:
Yep, works as expected!

openjck (approved):

PR #495, opened 2020-05-19
clean up global.css

spasovski:
Road to success:

1. This PR.
2. Extract core/color/spacing/typography eliminating the need of global.css 
3. Restructure Graph Paper to follow this same model.

This PR is a working concept and open to debate. Right now it simply moves as much of the CSS to its own svelte components (as quickly as possible) and cleans up some unused rules. The end goal is something like:

- core.css
- typography.css
- colorscales.css
- spacing.css

and the browser reading a single bundle.*hash*.css


src/components/explore/ComparisonSummary.svelte
```diff
@@ -46,152 +46,166 @@ $: displayValues = createNewPercentiles(left, right, keySet);
...
       <tr>
         <th class="value-label">{binLabel}</th>
-            {#if showLeft}
+          {#if showLeft}
             <th class:hidden={!hovered} class="summary-label ref">
-                  <div>
-                    <span class='small-shape'>●</span> {leftLabel || ''}
-                  </div>
+              <div>
+                <span class="small-shape">●</span> {leftLabel || ''}
```
hamilton, line 163: having a lot of trouble following the changes in this file and elsewhere. It looks like nothing changed except my editor is not replacing single quotes w/ doubles and some indent rules. Is this a lint rule that I'm not following along w/?

src/components/home/MarketingBlock.svelte
```diff
@@ -8,38 +8,42 @@ $: $t = 1;
...
-  margin-bottom: var(--space-4x);
-}
-
-.marketing-block__ask-us {
-  font-weight: 200;
-  font-style: italic;
-}
+  .marketing-block {
+    padding: var(--space-2x);
+      background: linear-gradient(to top, white, transparent), url('bg.jpg');
```
hamilton, line 13: confused by this – looks like this has the same indent rule as the original but it's just got an additional indent

src/udgl/layout/ContentBody.svelte
```diff
@@ -1,3 +1,11 @@
-<main class=content-content>
+<style>
+  .content-body-content {
```
hamilton, line 2: we can change this to just be `.content-body` since it's now component-scoped

hamilton: alternatively, you can ALSO just target `main` here and remove the class altogether. IMO `main` tells us semantically what the content is better than `content-body`. The rest is just style detail.

hamilton: Going with this approach would mean that you can remove a lot of classes throughout these components.

src/udgl/layout/ContentHeader.svelte
```diff
@@ -1,3 +1,16 @@
...
+    display: grid;
+    align-items: center;
+    height: 100%;
+    grid-auto-flow: column;
+    align-items: center;
+    justify-items: start;
+  }
+</style>
+
+<div class="content-header">
```
hamilton, line 14: you know, this element should be an actual `header` probably.

public/static/global.css
```diff
@@ -347,115 +348,9 @@ button:focus {
...
-  grid-column: body-content;
-  grid-row: body-footer;
-  display: grid;
-  align-items: center;
-  height: 100%;
-}
-
 /* surfaces are the main thing */
 /* needs depth and roundedness as the main concerns */
+/* davor note: these are used in stories only, they might get moved */
```
hamilton, line 353: we shoudl add a GH issue around updating storybook to have the same setup as graph-paper (including custom CSS)

hamilton (commented):
This is looking good & feels like its in the right direction @spasovski. I think we should probably chat sometime this week about the PR over video to talk through details and design decisions. Left some comments in the meantime. Reach out over Slack when you get a chance and we'll put some time on the calendar

src/components/explore/ComparisonSummary.svelte
openjck: It looks like the file was formatted with Prettier. Landing #480 will help. For now, there's an option in the *Files* tab for ignoring whitespace changes, which helps a lot.

openjck:
From browsing around the site, this looks good to me. The only difference I notice is that the *Histogram* block is bigger in search results, maybe as a result of the box-sizing change, but it still looks great.

src/components/explore/ComparisonSummary.svelte
hamilton: I see thanks for pointing that out

openjck:
Superseded by #609

PR #512, opened 2020-05-21
Fix #216: Use OptionMenu for large category sets

### LINKED ISSUE #216: come up w/ a way to handle large amounts of categories
There are some enumerated probes with a _ton_ of entries, most of which are zero. It would be great if we hid the long tail of nothing behind a "more..." button. This could be 100% local state.

Screen Shot 2020-01-23 at 7 31 14 AM

openjck:


src/components/explore/ProportionExplorerView.svelte
```diff
@@ -95,12 +92,37 @@ h2 {
...
+      {#if bucketOptions.length > maxVisibleCategories}
+        <OptionMenu multi on:selection={(evt) => {
+          dispatch('selection', { selection: evt.detail.keys, type: 'activeBuckets' });
+        }}>
+          {#each bucketOptions as bucketOption, i (bucketOption)}
+            <Option
+              selected={activeBuckets.includes(bucketOption)}
+              key={bucketOption}
+              label={bucketOption}>
+              <!--
```
hamilton: this note feels unnecessary in the source. This is a known design choice in svelte at this juncture and we follow this pattern throughout GLAM and graph-paper already

src/components/explore/ProportionExplorerView.svelte
```diff
@@ -95,12 +92,37 @@ h2 {
...
+                It's not currently possibly to specify a named slot for a
+                component, so we need to wrap this in a div.
+
+                https://github.com/sveltejs/svelte/issues/1037
+              -->
+              <div slot="right">
+                <ColorSwatch
+                  border="true"
```
hamilton: the black border makes this a bit hard on the eyes. Could we remove it?

src/components/explore/ProportionExplorerView.svelte
```diff
@@ -95,12 +92,37 @@ h2 {
 
     <div class=body-control-set>
       <label class=body-control-set--label>Categories</label>
+      {#if bucketOptions.length > maxVisibleCategories}
+        <OptionMenu multi on:selection={(evt) => {
+          dispatch('selection', { selection: evt.detail.keys, type: 'activeBuckets' });
+        }}>
+          {#each bucketOptions as bucketOption, i (bucketOption)}
```
hamilton: the ordering here in cases of a large quantity of options is not very usable. I have to go really far down the list to find the default checked ones, when what I think I want is the "top buckets" (the ones with actual color values mapped to them) on the top, then a divider line, then the rest below them (in whatever order). Is there a way to change this ordering so the most common ones are on the top, or the selected ones are on the top and in order?

src/components/explore/ProportionExplorerView.svelte
```diff
@@ -95,12 +92,37 @@ h2 {
...
+              <!--
+                It's not currently possibly to specify a named slot for a
+                component, so we need to wrap this in a div.
+
+                https://github.com/sveltejs/svelte/issues/1037
+              -->
+              <div slot="right">
+                <ColorSwatch
+                  border="true"
+                  color={bucketColorMap(bucketOption)} />
```
hamilton: I think we should lighten up the default color swatch to be `var(--cool-gray-subtle)`. I suspect you put the black border because this blends in with the background when you mouse over the option menu. The function that generates this color map could also just use a lighter cool gray instead.

hamilton (changes requested):
Looking good @openjck – left a few comments to address.

I am noticing that the alignment of these menu items are a bit off. The icon & swatch align, but the number looks like it is not aligned with the rest. Is this something that is addressable here in GLAM? If not, then we should re-create this use-case in graph-paper in storybook to make sure it works as expected.

Screen Shot 2020-05-26 at 9 24 46 AM

There is a merge conflict here that will need to be addressed, but that should be straightforward.

src/components/explore/ProportionExplorerView.svelte
hamilton: Keeping it in the component itself is a nice idea, but I would also not use `black` here. `var(--cool-gray-900)` would be a better choice of a default.

src/components/explore/ProportionExplorerView.svelte
openjck: I'm not sure `var(--cool-gray-subtle)` is visible enough on a white background. I'll have to think about this.

openjck:
> I am noticing that the alignment of these menu items are a bit off.

That's a quick fix in graph-paper. I'll submit a PR to address that now. I don't think we need to block merging this PR on that, but we will get the alignment fixed for free once optionmenu is upgraded.

openjck:
Here's the graph-paper PR:

https://github.com/graph-paper-org/graph-paper/pull/148

src/udgl/data-graphics/utils/color-maps.js
```diff
@@ -20,7 +20,7 @@ export function createCatColorMap(options) {
   return function catColorMap(v) {
     const i = getID(v);
     if (i >= 10 || i === undefined) {
-      return 'var(--cool-gray-200)';
+      return 'var(--cool-gray-400)';
```
hamilton:
```suggestion
      return 'var(--cool-gray-300)';
```

hamilton:
Thanks for the changes. I made one suggestion about the color choice. 400 feels very dark, and 300 would probably suffice here.

Is there any wy we can have the colored buckets (which are the ones with more counts than the others) at the top of the `sortedInactiveBuckets` list? As it is now, I clicked on one then had to scroll far to find it again. I think this is a pretty quick change, since you can just keep the inactive buckets with color assignemnts toward the top of that list.

If those makee sense, feel free to make those changes and merge at your discretion. Thanks!

hamilton (changes requested):
suggested changes in previous comment

src/udgl/data-graphics/utils/color-maps.js
openjck: I'm a little concerned that this isn't enough contrast, but this is fine.

src/config.json
```diff
@@ -1,4 +1,5 @@
 {
+  "numHighlightedBuckets": 10,
```
openjck: Note to self: Use the config/ directory instead, maybe config/firefox-desktop.js

openjck: NTS: Actually, create a file config/shared.js

src/udgl/data-graphics/utils/color-maps.js
```diff
@@ -20,7 +20,7 @@ export function createCatColorMap(options) {
   return function catColorMap(v) {
     const i = getID(v);
     if (i >= 10 || i === undefined) {
-      return 'var(--cool-gray-200)';
+      return 'var(--cool-gray-300)';
```
hamilton, line 23: I've gone back and forth on this, but I'm noticing it is actually now pretty hard for my eyes to distinguish the this shade from some of the other colors.  I played with this a bit, and I think for GLAM this works well to make it stand out on hover and also feels just dark enough to be visible + is clearly discernable from the other colors
```suggestion
      return 'var(--cool-gray-100)';
```

openjck: I merged this, but this doesn't look great. 100 is barely visible over the white background.

openjck: 100 is nice, but I find that it's difficult for me to see with my typical eyesight.

openjck: 100 does look nice in a way, but I find that it's difficult to see against a white background.

PR #531, opened 2020-05-22
Optimize random probes API

robhudson:


glam/api/views.py
```diff
@@ -329,43 +329,32 @@ def probes(request):
...
-                )
-            except NotFound:
-                continue
-            if aggregations:
-                probes.append({"data": aggregations, "info": probe.info})
-            if n <= len(probes):
-                break
+    # Get a random list of 9 probes from the Desktop nightly table.
+    aggs = DesktopNightlyAggregationView.objects.raw("""
+        SELECT id, metric, histogram
```
hamilton, line 345: Do we not get total clients here?

hamilton: probably below, not here, but I am not getting all the info I got before from this.

glam/api/views.py
```diff
@@ -329,43 +329,32 @@ def probes(request):
...
+        LIMIT %s
+    """, [n])
+
+    for agg in aggs:
+        try:
+            probe = Probe.objects.get(info__name=agg.metric)
+        except Probe.DoesNotExist:
+            pass
+
+        probes.append({"data": agg.histogram, "info": probe.info})
```
hamilton, line 362: seems like maybe here is where we'd put the other metadata llike total client counts

hamilton: Actually, nevermind. I'm just going to get rid of this part.

glam/api/views.py
robhudson: @hamilton Yeah it's not all the same data. Let me know what else you'd like and we can add it back. That was part of the process I thought we'd go over. Basically I was trying to streamline it based on what I see on the home page.

PR #544, opened 2020-05-26
fixes bug introduced in recent PR

hamilton:
A recent PR that optimized `toProportions` actually breaks the frontend. This fixes the fix


openjck (commented):
r+

PR #558, opened 2020-05-28
[do not merge] Fenix transpose demo branch

hamilton:
This branch has a dozen small workarounds to show fenix data today. IT doesn't work across the board. It's very busted in parts. Going to use this work to identify what needs to change in the frontend / backend to support fenix.


openjck:
Superseded by #607

PR #559, opened 2020-05-29
fixes #541 add probe name to page title

### LINKED ISSUE #541: Update HTML title to include probe name when on a probe detail page

spasovski:
fixes #541

### DIFF
--- a/src/routing/wrappers/Probe.svelte
+++ b/src/routing/wrappers/Probe.svelte
@@ -34,6 +34,14 @@
   });
 </script>
 
+<svelte:head>
+  {#if $store.route.section === 'probe'}
+    <title>{$store.probeName} | GLAM</title>
+  {:else}
+    <title>GLAM: Glean Aggregated Metrics Explorer</title>
+  {/if}
+</svelte:head>
+
 {#await $dataset}
   <div class="graphic-body__content">
     <Spinner size={48} color={'var(--cool-gray-400)'} />


src/routing/wrappers/Probe.svelte
```diff
@@ -34,6 +34,14 @@
   });
 </script>
 
+<svelte:head>
+  {#if $store.probeName}
+    <title>GLAM: {$store.probeName}</title>
```
openjck: What are your thoughts on reversing this order? That is, something like:

```svelte
    <title>{$store.probeName} - GLAM</title>
```

or

```svelte
    <title>{$store.probeName} | GLAM</title>
```

It's a common pattern, maybe more common than what we have now, and I think it would be especially helpful in GLAM where the name of the probe is so important. I think the probe name plus the favicon should give the user all the context they need.

openjck (commented):
Unfortunately, `probeName` doesn't always get cleared out. For example:

1. Load the homepage
2. Click on a probe
3. Click the back button

The title will still show the name of the probe. One workaround here might be to check if `$store.route.section === 'probe'`.

src/routing/wrappers/Probe.svelte
hamilton: I like this suggestion @openjck. Also my tabs tend to get short so having the probe name is the best thing while scanning, and the favicon does the work of saying what thing it is

spasovski:
Great suggestions thanks! Updated.

openjck:
I'm just realizing this is in *Probe.js*. Do we need to check `$store.route.section === 'probe'` considering? If not, would this make more sense somewhere else, like *Layout.svelte* or *Routing.svelte*?

openjck:
Superseded by #606.

PR #560, opened 2020-05-30
Fix #554: Add Google Analytics support

### LINKED ISSUE #554: Add Google Analytics
I don't think this will be too hard, given the way that page.js works. Google also has good documentation on usage within SPAs, which I don't recall seeing the last time I tried this.

https://developers.google.com/analytics/devguides/collection/gtagjs/single-page-applications

openjck:


openjck:
To test this, I added a GLAM property in my personal GA account and used that tracking code.

I don't have a corporate account tracking code yet, but I'm working on it.

glam/settings.py
```diff
@@ -120,6 +120,8 @@ class Core(Configuration):
         ],
     }
 
+    GA_TRACKING_ID = os.getenv("GA_TRACKING_ID", "")
```
robhudson: Use the django-configurations `values` that is already imported for environment variables, please. 

https://django-configurations.readthedocs.io/en/stable/values/#value-class

public/index.html
```diff
@@ -1,6 +1,21 @@
...
+    <script
+      async
+      src="https://www.googletagmanager.com/gtag/js?id={{ GA_TRACKING_ID }}"
+    ></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag() {
+        dataLayer.push(arguments);
+      }
+      gtag('js', new Date());
```
spasovski, line 14: they don't seem to document this anywhere though it gets described as "send a page hit to GA"

PR #561, opened 2020-05-30
Issue #251, issue #237: Fix 36 ESLint problems

openjck:


stories/basics/Color01.svelte
```diff
@@ -10,13 +9,31 @@ const families = [
 ];
 
 const tintsAndShades = (color) => {
-  const vals = [100, 150, 200, 250, 300, 350, 400, 450, 500, 550, 600, 650, 700, 750, 800, 850, 900];
+  const vals = [
```
spasovski, line 12: heh why, i doubt this can be changed easily and it's fine just....why

spasovski (approved):
Looks good, there's some stuff that could benefit from arrow functions but that's not the role of an ESLint pass

openjck:
> Looks good, there's some stuff that could benefit from arrow functions but that's not the role of an ESLint pass

Great catch. The odd syntax you're referring to is due to a catch-22 in the airbnb rules. If arrow function syntax is used with the return value directly after the arrow, ESLint complains that the line is too long. If the return value is put inside curly braces, ESLint complains that the return value should be directly after the arrow. If the return value is directly after the arrow, ESLint complains that the line is too long. And so on.

The workaround is to use parens.

See:
https://github.com/airbnb/javascript/issues/1532

stories/basics/Color01.svelte
openjck: We could change it in the future if we like. I'm not sure if there's a global option, but there's always single-line disables.

PR #566, opened 2020-06-01
Update development.md

hamilton:


docs/development.md
```diff
@@ -71,6 +71,17 @@ Data from glean may be pulled down using the following command:
 ./manage.py import_glean_aggs <PRODUCT>
 ```
 
+## Clearing your tables
+
+```
+make shell
+./manage.py dbshell
+DELETE FROM glam_desktop_nightly_aggregation;
+DELETE FROM view_glam_desktop_nightly_aggregation;
```
robhudson, line 80: You said you had problems with deleting data from the view so we can remove that. The view gets recreated after import anyway so removing that line would be fine.

robhudson:
Another option here would be to blow away the whole database docker image and start from scratch.

robhudson:
I wrote some docs that I'll submit soon outlining how to reset the docker container, which I think is simpler. So I'll close this out.

PR #587, opened 2020-06-08
counts now derived from histogram values as proportions + total_users

### LINKED ISSUE #578: in the API response, counts will need to be multiplied by the audienceSize
For categorical-like probes, the `counts` field used to be actual counts. Now, its a density, so it matches the `proportion` values. This means we'll have to multiply it by `audienceSize` in the right spot in `probe-utils.js` to get actual counts.

hamilton:
closes #578


openjck:
Superseded by #603

PR #594, opened 2020-06-09
basic search functionality from probe search service

spasovski:
This continues from hamilton's draft PR. Does not include updated styling considerations.

### DIFF
--- a/src/components/search/Search.svelte
+++ b/src/components/search/Search.svelte
@@ -1,41 +1,73 @@
 <script>
-  import { tick } from 'svelte';
+import { tick } from 'svelte';
 import { fly } from 'svelte/transition';
+import { cubicOut } from 'svelte/easing';
 import SearchIcon from 'udgl/icons/Search.svelte';
 import LineSegSpinner from 'udgl/LineSegSpinner.svelte';
-
-import {
-    store,
-} from '../../state/store';
-
-
-import telemetrySearch from '../../state/telemetry-search';
+import { throttle } from 'throttle-debounce';
 
 import TelemetrySearchResults from './SearchResults.svelte';
 
 
 let inputElement;
 let searchContainer;
+let results = [];
+let searchIsActive = false;
+let searchQuery = '';
+const SEARCH_THROTTLE_TIME = 500; // how often to send the PSS fetch (in ms)
 
 function turnOnSearch() {
-    store.setField('searchIsActive', true);
+  searchIsActive = true;
 }
 
 function unfocus() {
-    inputElement.blur();
-    store.setField('searchIsActive', false);
+  inputElement.blur();
+  searchIsActive = false;
 }
 
 async function onKeypress(event) {
-    if ($store.searchIsActive) {
-      const { key } = event;
-      if (key === 'Escape') {
-        await tick();
-        unfocus();
-      }
+  if (searchIsActive) {
+    const { key } = event;
+    if (key === 'Escape') {
+      await tick();
+      unfocus();
     }
+  }
 }
 
+const domain = (str, product = 'desktop') => {
+  const formattedStr = str.replace(' ', '%20');
+  return `https://dev.probe-search.nonprod.dataops.mozgcp.net/probes?limit=15&select=name,definition,type&product=eq.${product}&or=(name.eq.${formattedStr},index.phfts(english).${formattedStr},description.phfts(english).${formattedStr},name.ilike.${formattedStr}%)`;
+};
+
+function probeSearchServiceRequest(str) {
+  return results = fetch(domain(str)).then((r) => {
+    if (r.ok) return r.json();
+    return r;
+  });
+}
+
+let searchResults = Promise.resolve([]);
+let query = '';
+
+const handleSearchInput = throttle(SEARCH_THROTTLE_TIME, ({target: {value}}) => {
+  query = value;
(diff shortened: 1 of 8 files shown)


src/components/search/Search.svelte
```diff
@@ -1,41 +1,73 @@
...
+  if (searchIsActive) {
+    const { key } = event;
+    if (key === 'Escape') {
+      await tick();
+      unfocus();
     }
+  }
 }
 
+const domain = (str, product = 'desktop') => {
```
hamilton, line 38: just to note what we discussed over zoom: 

1. let's move this + the `probeSearchService` function over to `api.js`
2. let's decompose the `domain` function into its constituent parts so it is easier to reason about it

src/components/search/Search.svelte
```diff
@@ -1,41 +1,73 @@
...
+  const formattedStr = str.replace(' ', '%20');
+  return `https://dev.probe-search.nonprod.dataops.mozgcp.net/probes?limit=15&select=name,definition,type&product=eq.${product}&or=(name.eq.${formattedStr},index.phfts(english).${formattedStr},description.phfts(english).${formattedStr},name.ilike.${formattedStr}%)`;
```
hamilton, line 40: more follow-up from chatting:

(1) the `ilike` is technically not doing anything. Moving it to `name.ilike.*${formattedStr}*` will fix it
(2) adding `&type=neq.event` after the `product-eq` statement will filter out events, which I think we need here (we should not be returning events)

src/components/search/Search.svelte
```diff
@@ -101,47 +132,72 @@ async function onKeypress(event) {
...
-      value={$store.searchQuery} on:input={(evt) => {
-          store.setField('searchQuery', evt.target.value);
-          store.setField('searchIsActive', true);
-      }} />
-    </div>
+      value={searchQuery}
+      on:input={handleSearchInput}
+    />
+    {#if results.status}
+      <div
```
hamilton, line 184: I think I would expect the error message _in_ the search results area, not off to the side specifically. I would also expect it to be probably not yellow but kind of more in-line with our other error message.

src/components/search/SearchResults.svelte
```diff
@@ -195,56 +183,52 @@ li {
...
+      id="telemetry-search-results"
+      style="left: {x}px; width: {width}px;"
+      transition:fly={{ duration: 20, y: -10 }}
+      class="telemetry-results">
+        <div class="header-container">
+          {#if results.length}
+            <div class="header header--loaded" in:fly={{ x: -5, duration: 200 }}>
+              <div>found {results.length} probes</div>
+            </div>
           {:else}
```
hamilton, line 205: as this is now, I get a `getting the probes – one second!` when there are no search results. If not `results.length`, then we should be saying "no search results found for "____"`

src/components/search/SearchResults.svelte
```diff
@@ -195,56 +183,52 @@ li {
...
-          <div class="header" out:fly={{ x: 5, duration: 200 }}>
+            <div class="header" out:fly={{ x: 5, duration: 200 }}>
               <LineSegSpinner color={'var(--subhead-gray-02)'} />
-              <div>
-                  getting the probes – one second!
-              </div>
-          </div>
+              <div>getting the probes – one second!</div>
```
hamilton, line 208: I'm noticing that when I am waiting for the search service to return something, I'm actually not seeing any kind of spinner or anything in the title bar. We know when we're awaiting results – we could just set a flag somewhere in thsi file that is like `isCurrentlyFetching` and pop up the spinner on the right side of the search results header when that is true.

src/components/search/SearchResults.svelte
```diff
@@ -195,56 +183,52 @@ li {
...
-                  on:mouseover={() => { focusedItem = i; }}>
+                role="option"
+                id={name}
+                class:focused={focusedItem === i}
+                on:click={() => {
+                  page.show(`/firefox/probe/${results[focusedItem].name.toLowerCase()}/explore?${$currentQuery}`);
+                }}
+                on:mouseover={() => { focusedItem = i; }}>
                   <div class="name body-text--short-01">{name}</div>
                   <div class="probe-type label label-text--01 label--{type}">{type}</div>
```
hamilton, line 226: This seems a bit off. The label sizing for scalar is the same as it is for histogram, so the padding looks a little strange. I would expect the label to be right-aligned and not all the same width.

Screen Shot 2020-06-09 at 1 54 24 PM

hamilton (changes requested):
Thanks for filing @spasovski. I left some comments inline to get you started. They boil down to:

(1) let's filter out events from the postgrest query
(2) fix the `ilike` statement
(3) move the search API stuff into `api.js` as we discussed
(4) the probe type labels should be right-aligned and not identical width necessarily – is that a change from what's on dev?
(5) we need a little spinner at the top right of the search results header when we're awaiting a query to finish running. This can easily be done with a boolean that turns on / off while waiting for a search query
(6) if no results are present, we should say that instead of the "getting the probes – one second!" text
(7) the 400 error should probably be something in the search results body, not a yellow note to the right of everything

src/components/search/Search.svelte
robhudson: I would suggest making the base URL for the probe-search service to be configurable in some way. At some point we may have multiple instances for dev & prod.

openjck:
Superseded by #602

PR #599, opened 2020-06-10
Fix #598: Vertically-center quantile swatches

### LINKED ISSUE #598: color swatches no longer center-align in quantile button group
cc @openjck 

On stage, the color swatches are vertically centered with the text. On `master`, however, the swatches are aligned on the baseline, not centered.

Screen Shot 2020-06-09 at 8 56 59 PM

This issue also occurs iin proportion charts w < 10 categories.

openjck:
closes #598

### DIFF
--- a/src/components/controls/BodyControl.svelte
+++ b/src/components/controls/BodyControl.svelte
@@ -40,13 +40,21 @@ function toggle(v) {
 
 </script>
 
+<style>
+  .body-control__color-swatch-wrapper {
+    align-self: center;
+  }
+</style>
+
 <ButtonGroup {justify}>
   {#each options as {label, value, labelColor, tooltip, component}, i (label)}
     <Button tooltip={tooltip} level={level} compact={compact} toggled={multi
     ? selected.includes(value) : selected === value} on:click={() => { toggle(value); }
     }>
       {#if labelColor}
-        <ColorSwatch color={labelColor} />
+        <div class="body-control__color-swatch-wrapper">
+          <ColorSwatch color={labelColor} />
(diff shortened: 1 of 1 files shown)


openjck:
Putting this up for review in part because I'm not sure I understand how we name classes.

src/components/controls/BodyControl.svelte
```diff
@@ -40,13 +40,21 @@ function toggle(v) {
 
 </script>
 
+<style>
+  .body-control-color-swatch-wrapper {
```
hamilton: So we're using BEM (more or less), which means the class would be `.body-control__color-swatch-wrapper`. The `__` is because it is a child element.

Truthfully, I might just call this `.body-control--align-center`. The `--` signifies a modifier, (which is different than the element syntax `__`).  A single dash is just for delimiting words, so it isn't part of the convention.

We're not tied to this approach, but it does make it much easier to follow along nested CSS when we follow this convention and enforce some standards across our styles.

openjck: Sure thing. I'll use that name. I have to admit being skeptical of BEM, since my orientation has always been to avoid presentational class names, but I'm not familiar enough with the evidence for that approach to mount a big fight for it.

openjck:
Merging with Hamilton's blessing on Slack

src/components/controls/BodyControl.svelte
hamilton: Honestly, same. It's been fine and I weakly hold opinions about it. As long as we're consistent.

PR #596, opened 2020-06-09
moves the product params API & querystring parts to firefox-desktop.js

hamilton:
This is part of #569, closes #592. This moves `getParamsForQueryString` and `getParamsForDataAPI` to `firefox-desktop.js`.

PR #601, opened 2020-06-10
moves the product params API & querystring parts to firefox-desktop.js

openjck:
Supersedes #596 to merge into the *main* branch.

### DIFF
--- a/src/config/firefox-desktop.js
+++ b/src/config/firefox-desktop.js
@@ -46,6 +46,27 @@ export default {
       },
     },
   },
+  getParamsForQueryString(store) {
+    return {
+      channel: store.productDimensions.channel,
+      os: store.productDimensions.os,
+      aggregationLevel: store.productDimensions.aggregationLevel,
+      process: store.productDimensions.process,
+      timeHorizon: store.timeHorizon,
+      proportionMetricType: store.proportionMetricType,
+      activeBuckets: store.activeBuckets,
+      visiblePercentiles: store.visiblePercentiles,
+    };
+  },
+  getParamsforDataAPI(store) {
+    return {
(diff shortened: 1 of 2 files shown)


src/config/firefox-desktop.js
```diff
@@ -46,6 +46,27 @@ export default {
...
+      visiblePercentiles: store.visiblePercentiles,
+    };
+  },
+  getParamsforDataAPI(store) {
+    return {
+      channel: store.productDimensions.channel,
+      os: store.productDimensions.os,
+      probe: store.probeName,
+      process: store.productDimensions.process,
+      aggregationLevel: store.productDimensions.aggregationLevel,
```
robhudson, line 67: Can we add `product` here since the API supports it and would prefer to receive it to differentiate from fenix or other glean data... https://github.com/mozilla/glam/blob/main/glam/api/views.py#L303

robhudson: It's fine if the preference to wait. I'm just drive-by commenting. :)

hamilton: let's file an issue for this specifically. I think we're using `firefox` here which is probably not appropriate. We need to make sure we support the keys `firefoxDesktop` and `firefoxMobile` as two separate entities.

robhudson: The [constants.py](https://github.com/mozilla/glam/blob/main/glam/api/constants.py#L4-L7) file shows what I'm expecting in the API.

We can always change those keys. For mobile it may make sense to follow the app store IDs, e.g. `org.mozilla.fenix`.

spasovski:
Simple enough and looks good r+

PR #603, opened 2020-06-10
Fixes problem with counts in categorical explorer actually showing up as proportions

### LINKED ISSUE #578: in the API response, counts will need to be multiplied by the audienceSize

openjck:
closes #578

Supersedes #587 to merge into the *main* branch.


spasovski:
r+ code looks good. I can't speak to the strategy since I am not aware why it was broken before.

PR #694, opened 2020-06-29
fixes performance issue caused by tooltip updating in @graph-paper/core adding too many tasks to the queue

hamilton:
Could just use another pair of eyes locally to see if this fixes the issue.


spasovski (approved):
Looks great locally. Thanks!

PR #709, opened 2020-07-02
update front-end for probe-search changes

spasovski:


src/components/search/Search.svelte
```diff
@@ -1,6 +1,6 @@
 <script>
 import { tick } from 'svelte';
-import { fly, fade } from 'svelte/transition';
+import { fade } from 'svelte/transition';
```
robhudson, line 3: This PR is less fly

src/state/api.js
```diff
@@ -58,24 +58,34 @@ export async function getProbeData(params, token) {
...
+
     const params = new URLSearchParams();
     const queryOptions = [];
     const stringScalars =
-      '(type.eq.scalar,definition->details->>kind.eq.string)';
+      '(type.eq.scalar,info->calculated->latest_history->details->>kind.eq.string)';
 
     queryOptions.push(`name.eq.${str}`);
-    queryOptions.push(`index.phfts(english).${str}`);
+    queryOptions.push(`search.phfts(simple).${str}`);
```
robhudson: `search.plfts(simple)` -- It's the difference of a phrase query or plain query. I think plain will be better suited here.

src/components/search/Search.svelte
spasovski: Not the right time to fly these days

spasovski:
this got the go-ahead by @robhusdon via chat

PR #711, opened 2020-07-04
Fix #104: Persist reference point to URL

### LINKED ISSUE #104: add reference build id / version to querystring
Viewers will want to select a particular build / version and have that captured in the URL.

openjck:


spasovski (approved):
Code looks good 👍 works fine locally 👍 URL is even longer 😢

PR #722, opened 2020-07-10
Fix #687: Refactor store initialization

### LINKED ISSUE #687: Refactor GLAM's store instantiation to work like GUD's
Part of the Fenix story in #569

The store in GLAM has a lot of hard-coded functionality that gets in the way of generalizing for other products. One critical issue is the store instantiation on page load.

GUD has a nice way of instantiating the store that doesn't quite hard-code as much stuff. We could easily adapt GLAM's approach to follow that approach.

openjck:
Wanted to share my progress on this. This is untested, but I think it's a step in the right direction.

I'd like to merge some aspects of *firefox-desktop.js* into *shared.js* but may leave that for a later PR.

### DIFF
--- a/src/config/firefox-desktop.js
+++ b/src/config/firefox-desktop.js
@@ -21,6 +21,7 @@ export default {
         { key: 'Mac', label: 'Mac' },
         { key: 'Linux', label: 'Linux' },
       ],
+      defaultValue: 'Windows',
     },
     channel: {
       title: 'Channel',
@@ -30,6 +31,7 @@ export default {
         { key: 'beta', label: 'Beta' },
         { key: 'release', label: 'Release' },
       ],
+      defaultValue: 'nightly',
       isValidKey(key, probe) {
         return (probe.prerelease && key !== 'release') || !probe.prerelease;
       },
@@ -41,6 +43,7 @@ export default {
         { key: 'build_id', label: 'Build ID' },
         { key: 'version', label: 'Major Version' },
       ],
+      defaultValue: 'build_id',
     },
     process: {
       title: 'Process',
@@ -50,6 +53,7 @@ export default {
         { key: 'content', label: 'Content' },
         { key: 'gpu', label: 'GPU' },
       ],
+      defaultValue: 'content',
       isValidKey(key, probe) {
         return isSelectedProcessValid(probe.record_in_processes, key);
       },
--- a/src/config/shared.js
+++ b/src/config/shared.js
@@ -32,3 +32,20 @@ export function extractBucketMetadata(transformedData) {
   etc.bucketSortOrder = sorter;
   return etc;
 }
+
+export default {
+  activeBuckets: {
+    isMulti: true,
+    defaultValue: [],
+  },
+  proportionMetricType: {
+    defaultValue: 'proportions',
+  },
+  timeHorizon: {
+    defaultValue: 'MONTH',
+  },
+  visiblePercentiles: {
+    isMulti: true,
+    defaultValue: [95, 75, 50, 25, 5],
+  },
+};
--- a/src/state/store.js
+++ b/src/state/store.js
@@ -1,17 +1,82 @@
-import { derived } from 'svelte/store';
+import { derived, get } from 'svelte/store';
 
 import { createStore } from '../utils/create-store';
 
 // FIXME: take care of this dependency cycle.
 import { probeSet } from './telemetry-search'; // eslint-disable-line import/no-cycle
 
-import CONFIG from '../config/firefox-desktop';
+import sharedConfig from '../config/shared';
+import productConfig from '../config/products';
 
-// TODO: move this to the new config.js when 'product' is added.
-const DEFAULT_PROBE_PROCESS = 'content';
+export function getFromQueryString(fieldKey, isMulti = false) {
+  const params = new URLSearchParams(window.location.search);
+  const value = params.get(fieldKey);
+  if (isMulti) {
+    return JSON.parse(value);
+  }
+  return value;
+}
+
+function getDefaultState(
+  { basedOnQueryParams } = { basedOnQueryParams: false }
+) {
+  const state = {};
+
+  state.auth = {
+    isAuthenticated: false,
+    token: undefined,
+  };
+
+  // FIXME: applicationStatus or dashboardMode, not both.
+  state.applicationStatus = 'INITIALIZING';
+  state.dashboardMode = {};
+
+  state.product = window.location.pathname.split('/')[1] || 'firefox';
+  state.probeName = '';
+  state.recordedInProcesses = []; // Provided by the API. List of processes this probe was recording in.
+  state.reference = getFromQueryString('reference') || '';
+  state.route = {};
+
+  // Shared config
+  Object.entries(sharedConfig).forEach(([key, { isMulti, defaultValue }]) => {
+    if (basedOnQueryParams) {
+      state[key] = getFromQueryString(key, isMulti) || defaultValue;
+    } else {
+      state[key] = defaultValue;
+    }
+  });
+
+  // Product config
+  state.productDimensions = {};
+  Object.entries(productConfig[state.product].dimensions).forEach(
+    ([key, { defaultValue }]) => {
+      if (basedOnQueryParams) {
+        state.productDimensions[key] = getFromQueryString(key) || defaultValue;
+      } else {
+        state.productDimensions[key] = defaultValue;
+      }
+    }
+  );
+
+  return state;
+}
+
+export const store = createStore(getDefaultState({ basedOnQueryParams: true }));
+
+store.reset = () => {
+  store.reinitialize({
+    auth: store.getState().auth,
+    probeName: '',
+  });
+};
+
+function getActiveProductConfig() {
+  return productConfig[get(store).product];
+}
 
 export function getField(fieldKey) {
-  return CONFIG.dimensions[fieldKey];
+  const activeProductConfig = getActiveProductConfig();
+  return activeProductConfig.dimensions[fieldKey];
 }
 
 export function getFieldValues(fieldKey) {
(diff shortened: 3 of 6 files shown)


src/config/shared.js
```diff
@@ -32,3 +32,38 @@ export function extractBucketMetadata(transformedData) {
...
+    defaultValue: 'nightly',
+  },
+  os: {
+    defaultValue: 'Windows',
+  },
+  process: {
+    defaultValue: 'content',
+  },
+  product: {
+    defaultValue: 'firefoxDesktop',
```
robhudson: Just a caveat... the Glean name for Firefox will be `firefox.desktop`, if I saw that correctly. The API currently expects just `firefox`. I'm happy to change it to something that is more clearly "telemetry firefox" vs Glean firefox.

src/config/shared.js
```diff
@@ -32,3 +32,38 @@ export function extractBucketMetadata(transformedData) {
...
+    defaultValue: [],
+  },
+  aggregationLevel: {
+    defaultValue: 'build_id',
+  },
+  channel: {
+    defaultValue: 'nightly',
+  },
+  os: {
+    defaultValue: 'Windows',
```
robhudson: I've always felt the default here should be `*` for all OS's. And if the user wants to drill down they can. What do others think?

src/config/shared.js
```diff
@@ -32,3 +32,38 @@ export function extractBucketMetadata(transformedData) {
...
+    defaultValue: '',
+  },
+  timeHorizon: {
+    defaultValue: 'MONTH',
+  },
+  visiblePercentiles: {
+    isMulti: true,
+    defaultValue: [95, 75, 50, 25, 5],
+  },
+};
```
robhudson, line 51: Another item to possibly add here, or maybe it comes with a future code change, is `versions`, as in, the number of versions to query data for, defaulting to 3 if not sent to the API.

hamilton:
yeah, let's leave the moving of stuff from firefox desktop to shared for another PR.

src/config/shared.js
```diff
@@ -32,3 +32,38 @@ export function extractBucketMetadata(transformedData) {
   etc.bucketSortOrder = sorter;
   return etc;
 }
+
+export const facets = {
```
hamilton: Let's wait to do this for another PR. The default value for channels in other products are (for instance) not nightly. They're usually release. It'll be clearer what belongs in shared when we get our bearings, which will probably happen once people can see the data in GLAM.

src/config/firefox-desktop.js
```diff
@@ -12,47 +12,67 @@ import { validate, noResponse } from '../utils/data-validation';
...
-      title: 'Channel',
-      key: 'channel',
-      values: [
-        { key: 'nightly', label: 'Nightly' },
-        { key: 'beta', label: 'Beta' },
-        { key: 'release', label: 'Release' },
-      ],
-      isValidKey(key, probe) {
-        return (probe.prerelease && key !== 'release') || !probe.prerelease;
+    productDimensions: {
```
hamilton: Can you give more details why we have `dimensions.productDimensions` and not just `dimensions` here? This seems like an unnecessary nesting

src/state/store.js
```diff
@@ -5,21 +5,29 @@ import { createStore } from '../utils/create-store';
...
+  return pathProduct ? mappings[pathProduct] : defaultProduct;
+}
 
-// TODO: move this to the new config.js when 'product' is added.
-const DEFAULT_PROBE_PROCESS = 'content';
+const CONFIG = productConfig[getProduct()];
 
 export function getField(fieldKey) {
-  return CONFIG.dimensions[fieldKey];
+  return CONFIG.dimensions.productDimensions[fieldKey];
```
hamilton: These `getField` type functions are hard-coded to `CONFIG`, which means that whatever product is set on page load is what `CONFIG` resorts to. It seems to me that we should be using `getProduct` in each of these functions where appropriate. It's unclear how this will work with a different product as-is.

hamilton (changes requested):
This is looking good overall. Had two questions:

1. can you give more context for having `dimensions.productDimensions` vs. just `dimensions`? It's unclear what the additional nesting accomplishes.
2. there is more work to be done here re: initialization and how it will play with future products. Your `getProduct` function is a good start, but `CONFIG` is hard-coded to whatever product was there on page load, not future ones. Every single place where `CONFIG` is called should probably keep in mind what the _current_ product is.

src/state/store.js
```diff
@@ -5,21 +5,29 @@ import { createStore } from '../utils/create-store';
 // FIXME: take care of this dependency cycle.
 import { probeSet } from './telemetry-search'; // eslint-disable-line import/no-cycle
 
-import CONFIG from '../config/firefox-desktop';
+import productConfig from '../config/products';
+
+function getProduct() {
+  const defaultProduct = 'firefoxDesktop';
+  const mappings = {
```
hamilton: this mapping should probably live in `../config/products`, not here. That's the point of the config file.

hamilton: Also, we should remove the _need_ for this mapping. Are we not coding things properly b/t the API & GLAM? This should be a goal imo. Might be worth getting @robhudson's input here. If we have to ahve this mapping, that's fine, but would be worth filing an issue to fix this sooner rather than later so we don't have to maintain this.

hamilton (changes requested):
left out a few additional points

src/state/store.js
hamilton: Furthermore, we could have `getProduct` be a function in `productConfig` I think.

src/config/shared.js
openjck: >     1. can you give more context for having `dimensions.productDimensions` vs. just `dimensions`? It's unclear what the additional nesting accomplishes.

I'm fine removing that. It would make things easier. But I thought we agreed on Slack that we didn't want to flatten it out.

openjck:
> 1. can you give more context for having `dimensions.productDimensions` vs. just `dimensions`? It's unclear what the additional nesting accomplishes.

I'm fine with removing the `productDimensions` object. It would make things easier. But I thought we agreed on Slack that we didn't want to flatten it out.

> 2. there is more work to be done here re: initialization and how it will play with future products. Your `getProduct` function is a good start, but `CONFIG` is hard-coded to whatever product was there on page load, not future ones. Every single place where `CONFIG` is called should probably keep in mind what the _current_ product is.

I see. Yes, that's true. I'll look at that.

src/state/store.js
```diff
@@ -1,25 +1,91 @@
...
+    auth: store.getState().auth,
+    probeName: '',
+  });
+};
+
+const active = {
+  get productConfig() {
+    return productConfig[get(store).product];
+  },
+};
```
openjck: This allows us to do `active.productConfig.dimensions.[whatever]`. Another pattern I considered was adding a `get` method to the imported `productConfig` object, which would allow us to do `productConfig.currentProduct.dimensions.[whatever]`, which would look like this:

```javascript
Object.defineProperty(productConfig, 'currentProduct', {
  get() {
    return productConfig[get(store).product];
  },
});

const whatever = productConfig.currentProduct.dimensions.[whatever];
```

I'd be interested to hear which you think is more readable.

hamilton: I think the latter is a bit more readable to me b/c it's tied a bit more obviously to the config so I know where to look for the value if I see `productConfig.currentProduct` elsewhere. Also aren't we going to need `productConfig.currentProduct` elsewhere, such as in the dimension selector region? Is the goal to have other potential active things in `active`? If so, then I think this is a good approach, but I can't think of another case that we'd need this. My pref would be to go for `productConfig.currentProduct` with this in mind, but this is a weak position – either works for me.

src/config/shared.js
hamilton: Let's make sure to file these as follow-on issues.

src/state/store.js
```diff
@@ -1,25 +1,91 @@
...
+      } else {
+        state.productDimensions[key] = defaultValue;
+      }
+    }
+  );
+
+  return state;
+}
+
+export const store = createStore(getDefaultState());
```
hamilton: should this have `basedOnQueryParams: true`?

src/config/shared.js
```diff
@@ -32,3 +32,20 @@ export function extractBucketMetadata(transformedData) {
   etc.bucketSortOrder = sorter;
   return etc;
 }
+
+export default {
+  activeBuckets: {
```
hamilton, line 37: shouldn't `reference` also be here in some way? It doesn't look like the reference point persists now.

hamilton (changes requested):
This is looking great. I left a few comments about other things I'm noticing as I test out these changes:

1. some parameters don't seem to persist. It seems like that's probably due to `basedOnQueryParams` being `false` by default. Easy fix.
2. even if I set `basedOnQueryParams` manually on my local copy, it doesn't appear that the reference point persists from pageload to pageload. Probably worth taking a look at that, & I'm sure it's an easy fix.
3. there's a potential merge conflict with the changelog that'll need to be fixed before merging.

Other than these parts, I think this looks good. Feel free to address those and test / merge at your discretion. Thanks!

src/state/store.js
openjck: Sounds good. We may need `currentProduct` elsewhere, but I think *store.js* is mainly where things get weird: we need to read from the store itself to determine what `currentProduct` is.

I could add a `currentProduct` property to *config/products.js*. That would create a dependency cycle. I'm not certain that the dependency cycle would cause problems, but I try to avoid them in general.

src/config/shared.js
openjck: @robhudson Would you mind opening issues for this and the thing about firefox.desktop? I don't completely understand what you mean. I'll open an issue about OS defaults.

src/config/firefox-desktop.js
openjck: We hashed this out on Slack

src/state/store.js
openjck: I opted for something a little easier with less magic. May have been overthinking it.

src/state/store.js
openjck: Yes, important catch. Thank you!

src/config/shared.js
openjck: Would you mind opening a follow-up issue for this?

src/config/shared.js
openjck: I opened an issue for this: #753

PR #749, opened 2020-07-21
Add eslint-health script for CI

openjck:


openjck:
Note to self: Local and CI disagree about how many ESLint problems there are. I should fix that here.

PR #768, opened 2020-07-30
Update to include app IDs

robhudson:


robhudson:
Heads up to @hamilton @openjck @spasovski -- this PR will require both `product` and `app_id` (along with the usual dimensions) to be passed to the API for Fenix aggregation data.

glam/api/management/commands/import_glean_aggs.py
```diff
@@ -11,7 +12,11 @@
 
 
 GCS_BUCKET = "glam-dev-bespoke-nonprod-dataops-mozgcp-net"
-PRODUCT_MODEL_MAP = {"org_mozilla_fenix": "api.FenixAggregation"}
+APP_TO_MODEL = {
+    "org_mozilla_fenix": "api.FenixAggregation",  # nightly
+    "org_mozilla_firefox_beta": "api.FenixAggregation",  # beta
+    "org_mozilla_firefox": "api.FenixAggregation",  # release
```
acmiyaguchi, line 18: The `app_id`s in the probe scraper conventionally use kebab case: https://github.com/mozilla/probe-scraper/blob/master/repositories.yaml

acmiyaguchi (approved):
Looks reasonable to me, the only comment that the `app_id` is kebab cased in the probe scraper definition, and snake cased in the [etl scripts](https://github.com/mozilla/telemetry-airflow/blob/d13426a468f506567d9950a8c3635d08ed5db079/dags/glam_fenix.py#L19-L24). Being consistent with the ETL is probably the right choice here.

glam/api/management/commands/import_glean_aggs.py
robhudson: Thanks for pointing that out. The GLAM front-end will be passing this to the API, as well. I think the snake case is fine.

PR #781, opened 2020-08-04
PSS store.probe approach

spasovski:


openjck:
At the moment, this fails `npm run format:check` and `scripts/eslint-health`.

package.json
```diff
@@ -56,7 +56,6 @@
     "d3-scale-chromatic": "1.5.0",
     "d3-shape": "1.3.7",
     "d3-time-format": "2.2.3",
-    "flexsearch": "0.6.30",
```
openjck, line 59: :tada:

src/routing/Router.svelte
```diff
@@ -58,14 +59,20 @@
...
-        store.setField('probeName', probeName);
-        if ($probeSet) {
-          let newProbe = $probeSet.find((p) => p.name === probeName);
+        let newProbe;
+
+        // The canonical probe info fetch. (PSS)
+        getSearchResults(probeName).then((r) => {
+          newProbe = { ...r[0], loaded: true };
+          store.setField('probe', newProbe);
+          store.setField('probeName', probeName); // This is temporary: use store.probe.name
```
openjck: I'm not sure I understand this comment. Was this meant to be addressed before putting up the PR?

src/state/store.js
```diff
@@ -3,7 +3,7 @@ import { derived, get } from 'svelte/store';
 import { createStore } from '../utils/create-store';
 
 // FIXME: take care of this dependency cycle.
-import { probeSet } from './telemetry-search'; // eslint-disable-line import/no-cycle
+import probeSet from './probeset'; // eslint-disable-line import/no-cycle
```
openjck: It looks like we can remove the `eslint-disable-line` comment now. Hooray!

openjck:
I'm getting an error when selecting a result in search. If you're not seeing that error, we can hop on Zoom to reproduce it.

src/components/regions/ProbeDetails.svelte
```diff
@@ -49,7 +48,7 @@ function probeIsSelected(probeToTest) {
 </style>
 
 <div class="drawer details-drawer">
-{#if !$telemetrySearch.loaded}
+{#if !$store.probe.loaded}
```
hamilton, line 50: @spasovski would be good to go through this file and make sure we're pulling in alll the relevant metadata out of `$store.probe`. Could you make sure this file reaches parity w/ whatever was in $probe before?

spasovski: Yep I already had that in the initial PR should be easy.

src/routing/Router.svelte
spasovski: should be addressed before we merge to main yes, this PR is a draft

src/routing/pages/probe/FirefoxLegacyDetails.svelte
```diff
@@ -14,12 +14,12 @@
     event:
       'https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/collection/events.html',
     default:
-      'https://firefox-source-docs.mozi,l,la.org/toolkit/components/telemetry/collection/in,dex.html'
```
hamilton, line 17: what's this?

openjck: It works without the commas.

src/state/probeset.js
```diff
@@ -0,0 +1,15 @@
+import { readable } from 'svelte/store';
+
+// TODO: Make this dynamic based on prod vs local dev.
+const url = '__BASE_DOMAIN__/api/v1/probes/';
```
openjck, line 4: I know you didn't add this comment, but is it necessary? The code here seems fine; it's Rollup that needs to add a production domain.

src/routing/pages/probe/Table.svelte
```diff
@@ -4,14 +4,14 @@
...
-  import { store, probe } from '../../../state/store';
+  import { store } from '../../../state/store';
 </script>
 
 <Probe let:data let:probeType>
   <div class="graphic-body__content graphic-body__content--no-padding">
     <div in:fade>
       <!-- this conditional is a stopgap until we fix https://github.com/mozilla/glam/issues/206 -->
-      {#if $probe.type}
+      {#if $store.probe.loaded}
```
openjck, line 14: It looks like the issue mentioned in the comment has been resolved. Is this conditional still needed? If so, maybe we can update the comment explaining why.

src/routing/Router.svelte
```diff
@@ -47,25 +47,30 @@
 
   function useComponent(componentToUse, view) {
     return function handle({ params: { product, section, probeName } }) {
+      const storeValue = get(store);
```
openjck, line 50: Does `$store` not work in this case?

src/routing/Router.svelte
```diff
@@ -47,25 +47,30 @@
...
       // set it in the store and use store.resetProductDimensions()
       // to initialize.
-      if (product && $store.product !== product) {
+      if (product && storeValue.product !== product) {
         store.setField('product', product);
         store.resetProductDimensions();
       }
       // Issue #355: Update the probe here, whenever the path changes, to ensure
       // that clicks to the back/forward buttons work as expected.
+
```
openjck, line 61: Extra space. Not a big deal but sort of makes it less clear what the comment applies to.

src/routing/Router.svelte
```diff
@@ -47,25 +47,30 @@
...
       if (probeName) {
-        store.setField('probeName', probeName);
-        if ($probeSet) {
-          let newProbe = $probeSet.find((p) => p.name === probeName);
-          if (productConfig[$store.product].transformProbeForGLAM) {
-            newProbe = productConfig[$store.product].transformProbeForGLAM(newProbe);
+        // The canonical probe info fetch. (PSS)
+        getSearchResults(probeName).then((r) => {
+          let newProbe = { ...r[0], loaded: true };
```
openjck, line 67: Would it make sense to abstract this code into a little util function called something like `getProbeByName`?

src/components/table/DataTable.svelte
```diff
@@ -15,8 +15,8 @@ setContext('scrollLeft', scrollLeft);
 setContext('scrollTop', scrollTop);
 
 function updateScroll() {
-  $scrollLeft = container.scrollLeft;
-  $scrollTop = container.scrollTop;
+  scrollLeft.set(container.scrollLeft);
+  scrollTop.set(container.scrollTop);
```
openjck, line 19: Nice. I find this easier to read, assuming the top is valid at all, which I'm honestly forgetting right now.

hamilton: they're valid but if I were to  make an eslint set for svelte I would ban autosubscriptiions for setting. Just much easier to read!

openjck:
I'm noticing that the probe name doesn't appear in the `<title>`.

openjck:
A couple of things:

* If I visit the homepage, search for *gc_ms*, and then select the *gc_ms* probe, I'm brought to the explore page of the *gc_budget_ms* probe.
* It's not a big deal, but having seen it in practice, I'm not a huge fan of the console log since it crowds out other stuff. I wouldn't stress about it too much, but maybe in the long term we can use some kind of external logger like Sentry?

openjck:
I'm also wondering if it still makes sense to reset the store when the GLAM logo is clicked, now that we have routing.

If I visit the homepage, load a probe, press back, and then press forward, the load re-appears quickly. If I visit the homepage, load a probe, click the logo, and then press back, it takes a little while for the probe to reload.

openjck:
Some things I'm noticing:

- [x] The probe name doesn't appear in the `<title>`
- [x] If I visit the homepage, search for *gc_ms*, and then select the *gc_ms* probe, I'm brought to the explore page of the *gc_budget_ms* probe
- [x] It's not a big deal, but having seen it in practice, I'm not a huge fan of the console log since it crowds out other stuff. I wouldn't stress about it too much, but maybe in the long term we can use some kind of external logger like Sentry?
- [ ] I'm also wondering if it still makes sense to reset the store when the GLAM logo is clicked, now that we have routing. If I visit the homepage, load a probe, press back, and then press forward, the load re-appears quickly. If I visit the homepage, load a probe, click the logo, and then press back, it takes a little while for the probe to reload. (I probably should have looked at this when I implemented routing. If you'd like, I can help out on this. It may be better as a follow-up issue anyway.)
- [ ] If I load the homepage, search for *dns_renewal_time_for_ttl*, and then click it, I get a 404 error even though the data loads. This doesn't happen on main.

hamilton:
@openjck I think that your second item is an outstanding bug in the search functionality and should probably be filed as a separate issue if it is not already

src/routing/Router.svelte
hamilton: in hindsight I was trying to figure out if this issue was caused by autosubscriptions colliding with `page.js`. I think the source of the issue turned out to be elsewhere, so we could revert this.

openjck:
I opened #792 for the *dns_renewal_time_for_ttl* issue. I don't think we need to address the store resetting issue right now, if ever.

PR #786, opened 2020-08-05
Remove code and state after auth, add manual tests

openjck:


openjck:
Note that this splits the `tests/` directory into `tests/automated` and `tests/manual`. If that's too cute or annoying, just let me know and I'll move the acceptance tests somewhere else.

PR #790, opened 2020-08-06
Various authentication improvements

openjck:
* Upgrade the @auth0/auth0-spa-js Node module
* Remove the "code" and "state" query parameters from the URL after
  authentication completes
* Add manual acceptance tests for authentication


openjck:
Note that this doesn't fix the infinite redirect loop, but instead lays some groundwork that will be helpful when we work on it.

robhudson:
We decided to move to auth via nginx proxy. Closing this PR.

PR #791, opened 2020-08-07
Minor store cleanup

openjck:


src/state/store.js
```diff
@@ -76,12 +72,7 @@ export const store = createStore(getDefaultState({ basedOnQueryParams: true }));
 
 store.reset = () => {
   store.reinitialize({
-    auth: store.getState().auth,
-    probeName: '',
-    probe: {
-      name: '',
-      loaded: false,
-    },
```
openjck, line 82: This didn't occur to me earlier, but this shouldn't be necessary. The `reinitialize` method restores all of their values to their defaults, *except* things specified here. I refactored this a bit to make it more clear.

PR #871, opened 2020-09-08
Reset reference when navigating between probes

### LINKED ISSUE #850: reference store value does not reset between probe loads

robhudson:
Fixes #850


hamilton (approved):
Thanks @robhudson! Merging.

PR #888, opened 2020-09-15
Remove file system udgl package and use alias instead

robhudson:


wlach (approved):
Looks fine to me, seems simpler to use an alias than the previous approach.

PR #889, opened 2020-09-16
Add openresty auth proxy container

robotblake:


robhudson:
A couple general comments:

It would be cool if this were in its own repo and its own Docker container we could just add to our docker-compose and proxy the app through it. If glam is the first project to use this I could imagine it lives here short-term then moves to that kind of set up if it's successful. Is that the idea?

It would be good to see some docs showing how to spin this up and use it locally. In the no-auth branch I plan to add some docs for the docker commands to do a full production build locally and how to run the container to test it. I could see that being helpful here as well. That plus the existing docs on how to run it in dev mode during active development.

dockerfiles/proxy/Dockerfile
```diff
@@ -0,0 +1,17 @@
+ARG OPENRESTY_VERSION=1.17.8.2-4-buster-fat
+FROM openresty/openresty:${OPENRESTY_VERSION}
+
+ARG LUA_RESTY_OPENIDC_VERSION=1.7.3.1
+RUN opm get zmartzone/lua-resty-openidc=${LUA_RESTY_OPENIDC_VERSION}
+
+ENV NGINX_LISTEN=80
+ENV APP_LISTEN=127.0.0.1:8000
+ENV OIDC_REDIRECT_URI=/oidc/callback
+ENV OIDC_LOGOUT_PATH=/logout
```
robhudson, line 10: Are these pulled in as part of the `ngx.var.*` variables? They match the naming but the casing is different.

dockerfiles/proxy/access_by_openidc.lua
```diff
@@ -0,0 +1,102 @@
+local cjson = require("cjson")
+
+-- configs
+local opts = {
+    claim_groups = ngx.var.oidc_claim_groups or "https://sso.mozilla.com/claim/groups",
+    client_id = ngx.var.oidc_client_id,
+    discovery = ngx.var.oidc_discovery or "https://auth.mozilla.auth0.com/.well-known/openid-configuration",
+    logout_path = ngx.var.oidc_logout_path,
+    redirect_uri = ngx.var.oidc_redirect_uri,
+    scope = ngx.var.oidc_scope or "openid email profile",
```
robhudson, line 10: For the future I'm thinking `OIDC_SCOPE="email"` would be all we need to associate a Django user to an authenticated user.

Do you know what `profile` sends?

dockerfiles/proxy/Dockerfile
robotblake: These are inserted into the nginx config at https://github.com/mozilla/glam/blob/c4fb0bdc3ae93b1518d5dd28f7be2eb1838c224e/dockerfiles/proxy/default.conf.template#L61-L62 using `envsubst`, which then is accessed in the lua code via the `ngx.var.*` vars.

dockerfiles/proxy/access_by_openidc.lua
robotblake: This is more or less copied from what we use for other applications so I just left it as is. `openid` is required for doing the token exchange, and `profile` is used to get the users LDAP groups, which may not be needed for us at the moment but may be useful in the future.

I may need to potentially add the `read:aggregates` scope since we specified that when we set up the Auth0 creds initially.

robhudson: I think we can leave scopes out for now. If we need scopes in the future we can add something in.

glam/urls.py
```diff
@@ -1,29 +1,22 @@
...
     path("api/v1/data/", api_views.aggregations, name="v1-data"),
     path("api/v1/probes/", api_views.probes, name="v1-probes"),
     path("api/v1/probes/random/", api_views.random_probes, name="v1-random-probes"),
     path("api/v1/updates/", api_views.updates, name="v1-updates"),
-    path(
-        "api/v1/front-end-auth-config/",
-        auth_views.front_end_auth_config,
-        name="v1-front-end-auth-config",
-    ),
+    re_path(r"^(?!(?:api|static)(?:$|/)).*$", spa_view, name="catch-all"),
```
robhudson: If we just give a last pattern of `^.*/$` does that work? Whatever pattern doesn't match the above or Django apps would be caught by that and show the SPA view, rather than repeat this twice. I believe `static` should be part of that.

robotblake:
Adding the slash makes the root URL refuse to load, I tested with `r"^(?:.*/)?$"` which does appear to work. Maybe we can do some further digging into this tomorrow.

robhudson:
I did some more local testing and updated the URLs to use `path` to be simpler. Works great locally!

robotblake:
Note: We reverted the simpler `path` change because of a regression with direct-linking / refreshing.

PR #894, opened 2020-09-17
Add Fenix counts import script

robhudson:


glam/api/management/commands/import_glean_counts.py
```diff
@@ -0,0 +1,96 @@
...
+    def handle(self, *args, **options):
+
+        gcs_client = storage.Client()
+        bucket = gcs_client.get_bucket(GCS_BUCKET)
+
+        for product, opts in MAPPING.items():
+            model = apps.get_model(opts["model"])
+
+            for app_id in opts["apps"]:
+                filename = f"glam-extract-org_mozilla_fenix_glam_{app_id}-counts.csv"
```
acmiyaguchi, line 37: Technically the `app_id` would  be include the org_mozilla_fenix part:

```python
            for channel in opts["apps"]:
                app_id = f"org_mozilla_fenix_glam_{channel}"
                filename = f"glam-extract-{app_id}-counts.csv"
```

This breaks some of the other assumptions in `import_file`.

acmiyaguchi (approved):
This looks fine, although it does hardcode the `org_mozilla_fenix_glam` portion of the filename into the import. We currently only ingest fenix data, so this works. If we end up ingesting other glean data, then there will have to be smoe extra logic for fenix specific stuff.

robhudson:
Yeah, it's not perfectly set up to generically import any Glean products. I figured whatever product came next we can decide on and clean up the file naming conventions to be generic.

glam/api/management/commands/import_glean_counts.py
```diff
@@ -0,0 +1,96 @@
...
+                with connection.cursor() as cursor:
+                    cursor.execute(f"DROP TABLE {tmp_table}")
+                log(f"Deleting local file: {fp.name}.")
+                fp.close()
+
+    def import_file(self, tmp_table, fp, model, app_id):
+
+        csv_columns = [
+            f.name
+            for f in FenixCounts._meta.get_fields()
```
robhudson: `s/FenixCounts/model/`

PR #897, opened 2020-09-18
Update package.json scripts to not run sirv

robhudson:
We no longer need sirv since we have the nginx proxy set up via Docker.


robotblake (approved):
👍

PR #907, opened 2020-09-22
Attach labels to categorical histograms

robhudson:


src/config/firefox-desktop.js
```diff
@@ -112,11 +113,31 @@ export default {
...
+
+      // Attach labels to histogram if appropriate type.
+      if (metricType === 'histogram-categorical') {
+        const labels = {
+          ...appStore.getState().probe.info.calculated.latest_history.details
+            .labels,
+        };
+        data = produce(data, (draft) => {
+          draft.forEach((point, i) => {
+            draft[i].histogram = Object.entries(point.histogram).reduce(
```
wlach: I generally prefer to use `.map` instead of mutating things with`.forEach` where possible. This code may need some massaging for prettier, etc. but the general idea would be:

```suggestion
        data = produce(data, (draft) => {
          draft.map(point => {
            return {
                ...point,
                histogram: Object.entries(point.histogram).reduce(
```

wlach (approved):
I'm not super familiar with this code area, but this looks reasonable! I have a minor suggestion on forEach vs. map.

src/config/firefox-desktop.js
robhudson: That looks great, thanks!

Issue #716, opened 2020-07-07
change data time range to request last X versions


spasovski:
The API is ready. The UI needs to format the request in a way that sends the number of versions back we're trying to show.

robhudson:
Fixed in a579faa1020cf6532dd55e3df285b7c217445924

PR #928, opened 2020-10-01
Update time horizons to show more data (#716)

robhudson:
I consider this a working, work in progress.

TODO:
- [ ] Investigate triggering a re-fetch when the time horizon changes to > MONTH, and default back to a smaller number of versions requested from the API to keep the data response smaller by default
- [ ] Investigate making the API align more with the UI by request a number of builds rather than number of versions


robhudson:
Talked to hamillton and this is good enough for now. We agreed adjusting the API or re-querying the API may be better, but fetching 10 versions and slicing the data via the time horizon buttons is an good fix for now.

PR #932, opened 2020-10-02
Fix title not updating after selecting searched probe

### LINKED ISSUE #926: Switching probes doesn't update the head>title
STR:
- Open glam
- Type in search to find probe A, navigate to it
- From the probe detail page, type in probe to find probe B, navigate to it

Probe B's name doesn't get updated in the browser title.

robhudson:
Fixes #926

### DIFF
--- a/src/routing/wrappers/Layout.svelte
+++ b/src/routing/wrappers/Layout.svelte
@@ -13,6 +13,12 @@
   import routes from '../../config/routes';
 
   $: isProbeDetailsView = $store.route.section === 'probe';
+
+  const defaultTitle = 'GLAM: Glean Aggregated Metrics Explorer';
+  $: title =
+    $store.route.section === 'probe'
+      ? `${$store.probeName} | GLAM`
+      : defaultTitle;
 </script>
 
 <style>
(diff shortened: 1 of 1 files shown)


src/routing/wrappers/Layout.svelte
```diff
@@ -13,6 +13,15 @@
...
 
   $: isProbeDetailsView = $store.route.section === 'probe';
+
+  const defaultTitle = 'GLAM: Glean Aggregated Metrics Explorer';
+  $: title =
+    $store.route.section === 'probe'
+      ? `${$store.probeName} | GLAM`
+      : defaultTitle;
+  $: {
+    document.title = title;
```
wlach: So I'm not a svelte expert, but is this necessary? Shouldn't the `svelte:head` part set document.title?

wlach (changes requested):
Looks reasonable, one question which may be just my ignorance of what Svelte is doing.

src/routing/wrappers/Layout.svelte
robhudson: It should, yes. Let me test without that and see. I added that before I move the title changes to the script tag after learning that svelte has issues with updating the `<head>` inside an `{#if ...}` clause.

robhudson: Updated to remove that. It works just fine without it. Thanks!

wlach (approved):

Issue #634, opened 2020-06-16
Provide way to explore a query in greater detail


openjck:
This came up in our user interview with Eric. He mentioned that it might be interesting to provide a way to explore data in greater depth in BigQuery and/or STMO. Another option is to provide a textual BigQuery query which could be copied and pasted.

PR #950, opened 2020-10-07
Add ADR for data exploration options

robhudson:


docs/adr/0002-data-exploration.md
```diff
@@ -0,0 +1,52 @@
...
+Currently we provide the JSON representation of the data being displayed as an
+"Export to JSON" link
+
+Extra options we can provide:
+
+1. We could provide other alternate data file formats (e.g. CSV)
+2. We could provide the SQL to query the current set of data being displayed to
+   the user which can then be copy/pasted into Redash or BQ console as a
+   starting point for further exploration
+3. We could link directly to Redash with the query pre-populated
```
rafrombrc, line 31: Option 3 is pretty much a non-starter... given the BI tool evaluation that is currently happening, we don't know what role Redash will serve going forward, investing in a tighter integration with Redash would be unwise right now.

docs/adr/0002-data-exploration.md
```diff
@@ -0,0 +1,52 @@
...
+  but also help enable non-generalized use cases?
+
+## Considered Options
+
+Currently we provide the JSON representation of the data being displayed as an
+"Export to JSON" link
+
+Extra options we can provide:
+
+1. We could provide other alternate data file formats (e.g. CSV)
```
rafrombrc, line 27: This could make sense, but IMO only if we identify a concrete workflow that we're trying to serve, which would require some user research / interviews to find out. Are there specific product managers who would like to load the data into Excel to do further crunching? If yes, then adding CSV export sounds great! But we probably shouldn't build it unless we know there's a demand for the format(s) we're adding.

docs/adr/0002-data-exploration.md
```diff
@@ -0,0 +1,52 @@
...
+2. We could provide the SQL to query the current set of data being displayed to
```
rafrombrc, line 28: This seems to me the most useful initial deliverable.

robhudson: Also discovered this today: https://github.com/mozilla/glam/issues/634

PR #966, opened 2020-10-15
Fix #959: Make extracting bug IDs more robust

### LINKED ISSUE #959: Fenix: Support Github issue links in metric details sidebar
Currently the links for bugs and data reviews assume a bugzilla URL. We should support both bugzilla and github style URLs so we provide an accurate link name for the links.

robhudson:
Before | After
------ | -----
Screenshot_2020-10-16 network_cache_hit_time GLAM(1) | Screenshot_2020-10-16 network_cache_hit_time GLAM
Screenshot_2020-10-16 storage_stats_cache_bytes GLAM(2) | Screenshot_2020-10-16 storage_stats_cache_bytes GLAM(1)

### DIFF
--- a/src/utils/urls.js
+++ b/src/utils/urls.js
@@ -1,0 +1,16 @@
+export function extractBugId(url) {
+  if (url.includes('bugzilla')) {
+    const match = /id=(?<id>\d+)/.exec(url);
+    if (match) {
+      return `bugzil.la/${match.groups.id}`;
+    }
+  }
+  if (url.includes('github')) {
+    const regexp = /github\.com\/(?<org>[^/]+)\/(?<project>[^/]+)\/(pull|issues)\/(?<id>\d+)/;
+    const match = regexp.exec(url);
+    if (match) {
+      return `${match.groups.org}/${match.groups.project}#${match.groups.id}`;
+    }
+  }
+  return url;
+}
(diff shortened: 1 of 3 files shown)


src/utils/urls.js
```diff
@@ -0,0 +1,17 @@
...
+    const match = /id=(\d+)/.exec(url);
+    if (match) {
+      return match[1];
+    }
+    return url;
+  }
+  if (url.includes('github')) {
+    const match = /(issues|pull)\/(\d+)/.exec(url);
+    if (match) {
+      return `#${match[2]}`;
```
wlach: I think returning something like `mozilla/fenix#1234` would be better for GitHub issues (more idiomatic).

src/utils/urls.js
```diff
@@ -0,0 +1,17 @@
...
+    }
+    return url;
```
wlach: I don't think this is necessary.

```suggestion
```

wlach (changes requested):
One suggested change.

src/utils/urls.js
robhudson: This is if it's a link that isn't bugzilla or github so it doesn't show the link as `undefined`, just in case.

src/utils/urls.js
robhudson: This makes the links long and breaks the UI, but I can make each of those on their own row to allow for longer URL text...

Screenshot_2020-10-16 storage_stats_cache_bytes GLAM

PR #1021, opened 2020-11-10
Fix #634: Provide a way to explore data via SQL query

### LINKED ISSUE #634: Provide way to explore a query in greater detail

robhudson:
Screenshot_2020-11-10 gc_budget_overrun GLAM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [unreleased](https://github.com/mozilla/glam/compare/2020.11.0...HEAD) (date TBD)
 
+- Provide SQL query for further data exploration
+  ([#1021](https://github.com/mozilla/glam/pull/1021))
 - Remove default values from URLs so they aren't as unwieldy
   ([#1027](https://github.com/mozilla/glam/issues/1027))
 - Default to "All" ping types for unknown ping types
--- a/src/components/Modal.svelte
+++ b/src/components/Modal.svelte
@@ -1,0 +1,104 @@
+<script>
+  import { fade } from 'svelte/transition';
+  import { Close } from '@graph-paper/icons';
+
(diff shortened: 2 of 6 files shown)


wlach:
I have a few concerns about this:

1. (trivial) I don't think it will actually run, since we don't have a submission date filter?
2. (more serious) I think we actually need to provide a starting point that people will find useful, rather than just a view of the raw data. 

As is, I think this is just going to frustrate people, since in practice they'd need to go to #data-help to figure out what to do next (which might wind up being a very long and drawn out conversation). Earlier this year I wrote up this https://docs.telemetry.mozilla.org/cookbooks/main_ping_exponential_histograms.html which attempts to do this style of query on the main ping directly. Maybe we need to do an updated one that uses the GLAM tables and link to that as part of this interface? @acmiyaguchi, do you have any thoughts here?

robhudson:
> (trivial) I don't think it will actually run, since we don't have a submission date filter?

They do run, I've tested them. There does need to be a bit of context, however, since the glam database has more history whereas the ETL tables only have current version plus 2 versions prior. So the results could have less data when compared to glam.

wlach (approved):
We discussed this a bit on Slack, and think this looks pretty reasonable as a starting point for feedback on other people on what they'd like to see. In that case, maybe the best thing to do is land it and then ask people directly for feedback. We can then adjust the messaging, add docs, etc., etc.

The code itself looks good to me, I don't really have any feedback at this stage. 👍 

P.S. for future reference some results from this class of query are here:

https://sql.telemetry.mozilla.org/queries/76238/source

PR #1162, opened 2021-02-04
Fix #1143: Add SQL example using main_1pct

### LINKED ISSUE #1143: Add 2nd SQL query using main_1pct table
GLAM can fill in as much of the details of the query shown on the docs for greater exploration and customization of the aggregate columns...

https://docs.telemetry.mozilla.org/cookbooks/main_ping_exponential_histograms.html#slicing-along-arbitrary-dimensions

robhudson:
Updated modal with telemetry SQL tab selected...
Screenshot_2021-02-09 tls_delegated_credentials_time_until_ready_ms GLAM(1)


src/components/SqlModal.svelte
```diff
@@ -28,6 +22,142 @@
...
+      }
+    } else if ($store.probe.type === 'scalar') {
+      if (process === 'main') {
+        telemetryPath = `payload.processes.parent.scalars.${$store.probe.name}`;
+      } else {
+        telemetryPath = `payload.processes.${process}.scalars.${$store.probe.name}`;
+      }
+    }
+    const sql = [
+      'WITH per_build_client_day AS (',
```
wlach: I think we should find a way of storing this sql as a standalone template file of some kind. You should be able to do this with rollup-plugin-string: https://www.npmjs.com/package/rollup-plugin-string

src/components/SqlModal.svelte
```diff
@@ -28,6 +22,142 @@
...
+      '      udf.histogram_merge(',
+      '        ARRAY_AGG(',
+      '          udf.json_extract_histogram(',
+      `            ${telemetryPath}`,
+      '          )',
+      '        )',
+      '      )',
+      `  ) AS ${$store.probe.name}`,
+      '  FROM',
+      '    telemetry.main_1pct',
```
wlach: I think we should use `telemetry.main_nightly` if the channel is nightly, 1pct is very sparse for nightly leading to weird results.

Can you also add a SQL comment somewhere with a link to the docs we have on these tables:

https://docs.telemetry.mozilla.org/datasets/main_ping_tables.html

wlach (changes requested):
This is great stuff.

Can you link to [the cookbook](https://docs.telemetry.mozilla.org/cookbooks/main_ping_exponential_histograms.html) somewhere in the modal? I don't think we should assume it's generally known (I keep on telling people about it).

src/components/SqlModal.svelte
robhudson: Great idea. I can add one at the top of the SQL.

src/components/SqlModal.svelte
```diff
@@ -28,6 +25,75 @@
...
+      $store.productDimensions.channel === 'nightly'
+        ? 'main_nightly'
+        : 'main_1pct';
+    if ($store.probe.type === 'histogram') {
+      if (['main', 'parent'].includes(process)) {
+        telemetryPath = `payload.histograms.${$store.probe.name}`;
+      } else {
+        telemetryPath = `payload.processes.${process}.histograms.${$store.probe.name}`;
+      }
+    } else if ($store.probe.type === 'scalar') {
```
wlach: I don't think we want this path? As we discussed the query won't work with scalars.

src/stringTemplates/desktop-glam.tpl
```diff
@@ -0,0 +1,21 @@
...
+  total_users,
+  mozfun.glam.histogram_cast_json(aggregates) AS aggregates
+FROM
+  `moz-fx-data-shared-prod.telemetry.client_probe_counts`
+WHERE
+  metric="${ metric }"
+  AND channel="${ channel }"
+  AND ${ osFilter }
+  AND process="${ process }"
+  AND ${ buildIdFilter }
```
wlach:
```suggestion
  AND ${ buildIdFilter }

```

src/stringTemplates/desktop-telemetry.tpl
```diff
@@ -0,0 +1,67 @@
...
+    as_struct
+)
+SELECT
+  build_id,
+  percentile,
+  value
+FROM
+  percentiles
+CROSS JOIN
+  UNNEST(percentiles.percentile_nested);
```
wlach:
```suggestion
  UNNEST(percentiles.percentile_nested);

```

src/components/SqlModal.svelte
```diff
@@ -74,23 +183,51 @@
...
-    <button on:click={copyTextarea}>Copy to clipboard</button>
-    {#if status}
-      <span contenteditable="true" bind:textContent={status} transition:fade />
+    {#if activeTab === 2}
+      <p>
+        Read <a
+          href="https://docs.telemetry.mozilla.org/cookbooks/main_ping_exponential_histograms.html">
+          Visualizing Percentiles of a Main Ping Exponential Histogram</a>
+        for more on the use of this query.
+      </p>
```
wlach: 👍

wlach (changes requested):
In general looks good! Can you show some screenshots as well?

src/components/SqlModal.svelte
robhudson: Good catch, thanks.

robhudson:
Squashed and added screenshot

src/components/SqlModal.svelte
wlach: Looking at the screenshot, I'm wondering if it would look better if we pasted this text into the SQL statement as a comment. It looks a little weird to have this just below the `... :`

wlach (changes requested):
Sorry for the nitpicking, but what do you think about moving the copy/paste icon into the text area? Similar to what's done on dtmo:

image
https://docs.telemetry.mozilla.org/cookbooks/main_ping_exponential_histograms.html#getting-client-level-data

robhudson:
> Sorry for the nitpicking, but what do you think about moving the copy/paste icon into the text area? Similar to what's done on dtmo

I did consider that initially but it would mean refactoring the way I'm showing the SQL which is currently just in a textarea. But since you bring it up it sounds like that would be worthwhile to spend extra time on that. I changed it in this code from a button at the bottom since the telemetry SQL is so long and it is not longer visible at the bottom.

> It looks a little weird to have this just below the ... :

I could just get rid of the ellipsis and reword it so it's not necessary. I think the link works well there.

wlach:
> > Sorry for the nitpicking, but what do you think about moving the copy/paste icon into the text area? Similar to what's done on dtmo
> 
> I did consider that initially but it would mean refactoring the way I'm showing the SQL which is currently just in a textarea. But since you bring it up it sounds like that would be worthwhile to spend extra time on that. I changed it in this code from a button at the bottom since the telemetry SQL is so long and it is not longer visible at the bottom.

Not sure how hard it would be! If it's not too much trouble I would try that.

> > It looks a little weird to have this just below the ... :
> 
> I could just get rid of the ellipsis and reword it so it's not necessary. I think the link works well there.

That seems like a reasonable solution to me.

wlach (approved):

robhudson:
Thanks for all the reviews, @wlach!

PR #1222, opened 2021-03-01
Context menu

robhudson:


src/components/ContextMenu.svelte
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  function closeMenu() {
+    $showContextMenu = false;
+  }
+
+  // NOTE: I'd like to keep as much logic out of this file as much as possible
+  // and keep it focused on only displaying context menus. I'm not quite sure
+  // how to handle this with specific menu options and click event handlers,
+  // however.
```
wlach: Could you use a component binding? https://svelte.dev/tutorial/component-bindings

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -54,9 +62,34 @@
...
+
+  let menuPos = { x: 0, y: 0 };
+  let zoomUrl;
+  let hov;
+
+  function generateQueryString(paramsToUpdate) {
+    const activeProductConfig = getActiveProductConfig();
+    if (!activeProductConfig) return '';
+    const params = activeProductConfig.getParamsForQueryString($store);
+    return toQueryString({ ...params, ...paramsToUpdate });
```
wlach:
```suggestion
    return activeProductConfig ? toQueryString({ ...activeProductConfig.getParamsForQueryString($store), ...paramsToUpdate }) : '';
```

src/components/explore/ProbeExplorer.svelte
```diff
@@ -83,13 +83,29 @@
...
+      if (found) return found;
+    }
+    return data[data.length - 1];
+  }
+
+  let ref = getDefaultReferencePoint();
+
   function setDomain(str) {
-    if (aggregationLevel === 'build_id') {
+    if (str === 'ZOOM') {
```
wlach, line 98: I would personally use constants for these things. e.g. `GLAM_ZOOM_KEY` or whatever

src/components/explore/ProbeExplorer.svelte
```diff
@@ -83,13 +83,29 @@
...
+    if ($store.ref) {
+      const found = data.find((d) => d[aggregationLevel] === $store.ref);
+      if (found) return found;
+    }
+    return data[data.length - 1];
+  }
+
+  let ref = getDefaultReferencePoint();
+
   function setDomain(str) {
```
wlach, line 97: While you're here, could you use a more descriptive variable name?

```suggestion
  function setDomain(domain) {
```

wlach (changes requested):
This looks pretty reasonable in outline-- is there any way for me to test this? See my question in #glam-dev.

wlach:
Some additional feedback after trying this out:

* The right-click behaviour is not really discoverable. Some kind of brief text like *right click on hover to zoom* just below the graph would really help.
* It would be nice if the hovering over behaviour with the graph was disabled while the prompt (zoom from here to the ref point) is enabled
* It wasn't immediately clear to me how to reset the zoom when zoomed in. possibly you could do that by adding a "custom" setting to time horizon when zooming in, and animating the addition of that to the time horizon box: 
image

Other than that it seems to work as expected!

wlach:
> It wasn't immediately clear to me how to reset the zoom when zoomed in. possibly you could do that by adding a "custom" setting to time horizon when zooming in, and animating the addition of that to the time horizon box:

So to be clear in that case, the time horizon box would then look like this with the custom option (which would always be selected if present):

image

In my mind, clicking on any of the other options would hide the custom zoom option.

wlach:
BTW, I used a similar concept for mission control and I think it works pretty well: https://missioncontrol.telemetry.mozilla.org/#/release/mac/main_crashes

image

I believe GUD also does something like this.

robhudson:
Merged some fixes to get on dev for further testing. Incremental improvements will be added in other commits.

PR #1323, opened 2021-04-26
Sort key label order (fixes #1313)

### LINKED ISSUE #1313: Label order in dropdown seems random, prefer alphabetical
I have a `labeled_counter` metric for `perf_startup_startup_type` ([dashboard](https://glam.telemetry.mozilla.org/fenix/probe/perf_startup_startup_type/explore?ref=2021041717)). It has many labels and I opened the dropdown:

image

The labels appear in a random order that makes it difficult for me to find the one I want. I think it'd be preferable to me if the labels were in alphabetical order so it'd group related items (i.e. `warm_main`, `warm_unknown`, and `warm_view` would all be together).

Iinh:
Fixes #1313.

Tagging @robhudson for review. Btw could you add me as a collaborator to this repo? Thanks!

### DIFF
--- a/src/components/controls/ProbeKeySelector.svelte
+++ b/src/components/controls/ProbeKeySelector.svelte
@@ -10,6 +10,8 @@
   let button;
   let width;
 
+  options.sort();
+
   function toggle() {
     active = !active;
   }
@@ -56,7 +58,7 @@
       text: 'this probe has multiple keys associated with it',
     }}>
     <div>
-      {#each options as opt, i}
+      {#each options as opt}
         <div
           style="
             visibility: {opt === currentKey


robhudson:
I'm actually not seeing the cited example being sorted with this change. Here's a screenshot of the menu items...
Screenshot 2021-04-26 at 17-51-12 perf_startup_startup_type GLAM

Iinh:
Oh that's strange, it's sorted on mine.

CleanShot 2021-04-26 at 18 09 51@2x

Applying `sorted()` to the `options` array should sort the menu options. I'm not sure what's going on...

robhudson:
Okay, it's working for me this morning when I restarted my app. Apologies.

Should we apply the `sort` to the 2nd place the `options` is iterated below this one?

src/components/controls/ProbeKeySelector.svelte
```diff
@@ -4,6 +4,7 @@
   import { tooltip as tooltipAction } from '@graph-paper/core/actions';
 
   export let options;
+  options.sort();
```
Iinh: @robhudson since we want to sort options in 2 locations, do you think it's better that we just sort it here instead?

robhudson: If we need it sorted in both places? I wasn't sure what that 2nd loop actually did at first glance. But sorting it here in the JS portion of the svelte code works too. Whatever you think is best!

robhudson (approved):

PR #1327, opened 2021-04-28
Update link to Glean Dictionary to production URL

wlach:


Iinh (approved):
new production url yay!

Issue #1135, opened 2021-01-26
The summary table is cramped


mstange:
On [the page for the `content_frame_time_reason` probe](https://glam.telemetry.mozilla.org/firefox/probe/content_frame_time_reason/explore?activeBuckets=%5B%22OnTime%22%2C%22SlowComposite%22%2C%22MissedCompositeLong%22%2C%22MissedCompositeMid%22%2C%22MissedComposite%22%2C%22MissedCompositeLow%22%2C%22NoVsyncNoId%22%2C%22NoVsync%22%5D&process=parent&ref=20210115035053), the Summary table on the right doesn't have enough space for the reasons.

Many rows show "MissedC...".

<img width="1328" alt="Screen Shot 2021-01-26 at 12 33 33 PM" src="https://user-images.githubusercontent.com/961291/105881830-f863e800-5fd2-11eb-99ea-64ecca9c92fa.png">

Iinh:
Closed via #1339.

PR #1339, opened 2021-05-04
Show full words in summary table

Iinh:
Because we use ellipses for overflow content, it's hard to read what is in the column, as shown in #1135. This is fixed by removing overflow ellipses and and increasing the width of the column for more real estate.

**Before**

image

**After**

Screen Shot 2021-05-04 at 9 07 03 AM

### DIFF
--- a/src/components/explore/ComparisonSummary.svelte
+++ b/src/components/explore/ComparisonSummary.svelte
@@ -120,9 +120,7 @@
   .value-label {
     min-width: calc(var(--space-base) * 7);
     max-width: calc(var(--space-base) * 10);
-    white-space: nowrap;
-    overflow: hidden;
-    text-overflow: ellipsis;
+    overflow-wrap: break-word;
   }
 
   .value-change {


src/components/explore/ComparisonSummary.svelte
```diff
@@ -118,11 +118,9 @@
   }
 
   .value-label {
-    min-width: calc(var(--space-base) * 7);
+    min-width: calc(var(--space-base) * 9);
```
robhudson: Does the extra width here cause any problems? I didn't notice any in your screen shots but wondering why this is needed in particular?

robhudson (approved):

src/components/explore/ComparisonSummary.svelte
Iinh: yeah I wanted more space for the columns so we can see the categories more easily, but the difference is actually not that big so I decided to revert this.

robhudson: If you prefer it let's keep it. I was mostly curious. Otherwise feel free to merge.

PR #1342, opened 2021-05-04
Properly render markdown of probe description

Iinh:
Probe/metric description is written in markdown format, but since we're not rendering them as markdown it's causing some display issue, such as:

Screen Shot 2021-05-04 at 3 09 06 PM

Or this [funny case of browser rendering `<marque>` tag in description](https://dev.glam.nonprod.dataops.mozgcp.net/firefox/probe/use_counter2_onfinish_page/explore?activeBuckets=%5B%22no%22%2C%22yes%22%5D&process=parent):

**Before:**

**After:**

Screen Shot 2021-05-04 at 3 44 25 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "jest": "26.6.3",
     "jest-transform-svelte": "2.1.1",
     "lodash": "4.17.21",
+    "marked": "2.0.3",
     "npm-run-all": "4.1.5",
     "prettier": "2.2.1",
     "prettier-plugin-svelte": "2.2.0",
@@ -61,7 +62,6 @@
     "d3-scale-chromatic": "2.0.0",
     "d3-shape": "2.1.0",
     "d3-time-format": "3.0.0",
-    "marked": "2.0.1",
     "page": "1.11.6",
     "throttle-debounce": "3.0.1"
   },
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -1,0 +1,20 @@
+<script>
+  import { use, parse, parseInline } from 'marked';
+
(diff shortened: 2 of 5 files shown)


src/components/Markdown.svelte
```diff
@@ -0,0 +1,26 @@
+<script>
```
Iinh, line 1: (We use the same [Markdown component](https://github.com/mozilla/glean-dictionary/blob/main/src/components/Markdown.svelte) in Glean Dictionary.)

src/components/Markdown.svelte
```diff
@@ -0,0 +1,26 @@
...
+  import { use, parse, parseInline } from 'marked';
+
+  const renderer = {
+    list(body, ordered, start) {
+      const outerEl = ordered ? 'ol' : 'ul';
+      return `<${outerEl} ${
+        start ? `start="${start}"` : ''
+      } class="mzp-u-list-styled">${body}</${outerEl}>`;
```
robhudson: I don't think this class applies for glam.

robhudson (changes requested):

src/components/Markdown.svelte
Iinh: you're right! Thanks for catching that!

package.json
```diff
@@ -43,6 +43,7 @@
     "jest": "26.6.3",
     "jest-transform-svelte": "2.1.1",
     "lodash": "4.17.21",
+    "marked": "^2.0.3",
```
robhudson: We usually remove the `^` from the package.json file.

robhudson (approved):

PR #1347, opened 2021-05-05
Add Mozilla code of conduct

mreid-moz:
Add Mozilla code of conduct as described [here](https://wiki.mozilla.org/GitHub/Repository_Requirements).

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
 public/static/bundle.js
 public/static/bundle.css
 tests/data/
+CODE_OF_CONDUCT.md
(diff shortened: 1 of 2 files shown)


mreid-moz:
Oops, looks like the `prettier` check doesn't like the long lines in the code of conduct file.

Since it's a copy of [a file from elsewhere](https://github.com/mozilla/repo-templates/blob/master/templates/CODE_OF_CONDUCT.md), does adding it to the `.prettierignore` file sound reasonable?

robhudson:
> Since it's a copy of [a file from elsewhere](https://github.com/mozilla/repo-templates/blob/master/templates/CODE_OF_CONDUCT.md), does adding it to the `.prettierignore` file sound reasonable?

Yep that sounds good, we can ignore it.

robhudson (approved):

PR #1348, opened 2021-05-05
Fix #1124: Add ability to get pushlog for range

### LINKED ISSUE #1124: Ability to get a pushlog range from the history view
(Migrated from https://github.com/mozilla/telemetry-dashboard/issues/248 )

It should be possible to select a range in the history graph and obtain a URL for a pushlog ([example](https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=c4186bb32c302ec2c82b3b0323af828ccebe0b48&tochange=bd7c35b7e234712f94f6dbb755ad983c1db8b07b)) that lists the changes that landed in the selected range.

Motivation:

When looking at the historical graph of a telemetry probe, the main question on my mind is:
"Did anything change, and if so, why?"
The "why" is currently rather cumbersome to track down. There are [manual steps](https://github.com/mozilla/telemetry-dashboard/issues/248#issuecomment-728315593) to get there, but since this is such a central workflow, we should automate as much of the process as we can.

This probably needs to implemented separately per product. For Gecko + Firefox Desktop, you can get the revision from the build ID using a rest query to https://buildhub.moz.tools/ (@wlach knows the specifics) and then build the pushlog URL. For Fenix, the user probably wants to see a commit log on github.

robhudson:
To test this locally the following commands need to be run:
- `make shell` to get into the docker container
- `./manage.py migrate` to set up the new database table
- `./manage.py import_desktop_aggs nightly` to run the import


glam/api/management/commands/import_desktop_aggs.py
```diff
@@ -124,3 +128,52 @@ def import_file(self, tmp_table, fp, model, channel):
...
+            log(channel, "No SHAs to update")
+            return
+
+        query = """
+            SELECT build.build.id, build.source.revision
+            FROM `moz-fx-data-shared-prod.telemetry.buildhub2`
+            WHERE build.build.id IN UNNEST(@build_ids)
+            AND build.target.channel = @channel
+            GROUP BY 1, 2
+        """
```
Iinh, line 164: if I want to see the output of this query in redash, what should I replace `@build_ids` and `@channel` with?

Iinh, line 164: if I want to see the output of this query in redash, what should I replace `@build_ids` and `@channel` with?

glam/api/models.py
```diff
@@ -202,3 +202,19 @@ class Meta:
                 fields=["app_id", "channel", "version", "ping_type", "build_id", "os"],
             )
         ]
+
+
+class FirefoxBuildRevisions(models.Model):
```
Iinh, line 207: this class autogenerates the `FirefoxBuildRevisions` table in `0017_add_shas_table.py ` when we run `./manage.py migrate`?

src/components/ChartContextMenu.svelte
```diff
@@ -152,12 +154,18 @@
...
-          <i>Coming soon: PUSHLOG LINK</i>
+      {#if pushlogUrl}
+        <div class="option">
+          <div class="option-icon">
+            <a href={pushlogUrl} target="pushlog">
+              <GitBranch size="12" />
+            </a>
+          </div>
+          <div class="option-link">
+            <a href={pushlogUrl} target="pushlog">View Changesets in Range</a>
```
Iinh, line 165: What do you think about using a more definitive name like "View Pushlog Range" for this feature, instead of "View Changesets in Range"?

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -103,34 +105,44 @@
...
-        : hovered.datum.build_id;
-    clickedRef = getDefaultReferencePoint();
-    clickedRef =
-      aggregationLevel === 'build_id'
-        ? clickedRef.build_id
-        : clickedRef.version;
+    let referencePoint = getDefaultReferencePoint();
+
+    clickedHov = hovered.datum.build_id;
+    clickedRef = referencePoint.build_id;
```
Iinh, line 111: I noticed we used to assign `hovered.datum.build_id` to `hovered.datum.version` but removed them in this change, what did we use `hovered.datum.version` for?

Iinh (commented):
This looks good, I tested around with this feature locally and it worked great! I have some clarifying questions to help me understand the code better which I've left in the comments below. Thanks!

glam/api/management/commands/import_desktop_aggs.py
robhudson: You could run something like this to get similar results. The `build.build.id` WHERE clause just gets only the builds we ask for that we have imported and don't have revisions for yet...

```sql
SELECT
  build.build.id,
  build.source.revision
FROM
  `moz-fx-data-shared-prod.telemetry.buildhub2`
WHERE
  build.target.channel = "nightly"
GROUP BY
  1,2
LIMIT
  10
```

glam/api/models.py
robhudson: This Django model will be used to create a migration file when we run `./manage.py makemigrations`, which I did already and the resulting file is `glam/api/migrations/0017_add_shas_table.py`. Then when we run `./manage.py migrate` it reads this file and creates the database table (or whatever migration action is listed there).

src/components/explore/AggregationsOverTimeGraph.svelte
robhudson: A little above this we are (line 103) we are returning if the `aggregationLevel` is not `build_id`, so this extra code would never have been called anyway. I decided to clean it up and make it simpler.

src/components/ChartContextMenu.svelte
robhudson: That doesn't sound accurate to me. The link will take you to a page that lists all changesets between the two revisions (selected from the build ID range).

Iinh (approved):
Thank you for answering my questions!

PR #1349, opened 2021-05-06
Link to DTMO docs

Iinh:
Add the link to DTMO docs in footer.

Screen Shot 2021-05-06 at 10 13 05 AM

### DIFF
--- a/src/components/regions/Footer.svelte
+++ b/src/components/regions/Footer.svelte
@@ -13,11 +13,10 @@
   }
   .project-links {
     list-style: none;
-    display: grid;
+    display: flex;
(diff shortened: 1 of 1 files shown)


src/components/regions/Footer.svelte
```diff
@@ -13,11 +13,10 @@
   }
   .project-links {
     list-style: none;
-    display: grid;
+    display: flex;
```
Iinh, line 16: using flexbox instead of grid here because grid jumped line with 4 links.
Screen Shot 2021-05-06 at 10 10 51 AM

robhudson (approved):
Great suggestion!

PR #1367, opened 2021-05-12
Fix BodyControl component not showing (fixes #1364 and closes #1365)

### LINKED ISSUE #1364: Bring back the chart / table toggle
Previously there was a toggle to switch between the chart view and table view. Visually it was flush left of the various dimensions selectors under the logo. We aren't sure when or how they are no longer being displayed.

This issue is to investigate what happened and implement a fix to bring them back.

### LINKED ISSUE #1365: Bring back the percentiles toggles
Previously the user had the ability to enable or disable each percentile on the graph, but this has since disappeared. Visually this was just under the text "Probe Value Percentiles". Let's figure out what happened and bring this ability back.

Iinh:
In #1243 we [added the `enabled` property to `options`](https://github.com/mozilla/glam/pull/1243/files#diff-6484fee8af831fc6220645e6aee7e6443b02644c8d1a5009bf99fdcf124ed296) in `TimeHorizonControl`, but we didn't add it to the other `options` objects. This causes `BodyControl` -- the toggle component -- to dissapear because it relies on `options.enabled` to be true

https://github.com/mozilla/glam/blob/7d3c016e1186a66148a4ae1f18b2ce3144959ab1/src/components/controls/BodyControl.svelte#L50

Adding `options.enabled` back fixes #1364 and #1365.

### DIFF
--- a/src/components/controls/KeySelectionControl.svelte
+++ b/src/components/controls/KeySelectionControl.svelte
@@ -10,6 +10,7 @@
     label: opt,
     value: opt,
     labelColor: colorMap(opt),
+    enabled: true,
   }));
 </script>
 
--- a/src/components/controls/PercentileSelectionControl.svelte
+++ b/src/components/controls/PercentileSelectionControl.svelte
@@ -12,11 +12,35 @@
 
 <BodyControl
   options={[
-    { label: '5%', value: 5, labelColor: cmp(5), tooltip: t(5) },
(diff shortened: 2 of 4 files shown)


Iinh:
Screenshot

Screen Shot 2021-05-12 at 12 31 24 PM

robhudson (approved):

PR #1379, opened 2021-05-18
Rescale y-axis when percentiles are toggled

### LINKED ISSUE #627: Re-scale y-axis when percentiles are enabled and disabled
If percentiles are disabled, for example, and the remaining lines are squished at the bottom of the graph, we should re-scale the y-axis so that they are easier to read.

Iinh:
Fixes #627.

### DIFF
--- a/src/components/explore/AggregationsOverTimeGraph.svelte
+++ b/src/components/explore/AggregationsOverTimeGraph.svelte
@@ -143,6 +143,28 @@
     // Finally, set the flag to open the context menu.
     $showContextMenu = true; // eslint-disable-line no-unused-vars
   }
+
+  const getUniqueArray = (arr) => {
+    let a = arr.concat();
+    for (let i = 0; i < a.length; ++i) {
+      for (let j = i + 1; j < a.length; ++j) {
+        if (a[i] === a[j]) a.splice(j--, 1);
+      }
+    }
+    return a;
+  };
+
+  const getYDomain = (percentiles) => {
+    let percentileData = [];
+    for (const p in percentiles) {
+      percentileData = [
+        ...percentileData,
+        ...data.map((arr) => arr.percentiles[percentiles[p]]),
+      ];
+    }
+    return getUniqueArray(percentileData).sort((a, b) => a - b);
+  };
+  $: yValues = getYDomain($store.visiblePercentiles);
 </script>
 
 {#if showContextMenu}
@@ -164,7 +186,7 @@
   </ChartTitle>
   <DataGraphic
     {xDomain}
-    {yDomain}
+    yDomain={yValues}
     yType={yScaleType}
     xType={xScaleType}
     height={aggregationsOverTimeGraph.height}
--- a/src/components/explore/AggregationsOverTimeGraph.svelte
+++ b/src/components/explore/AggregationsOverTimeGraph.svelte
@@ -144,28 +144,31 @@
     return data[data.length - 1];
   }
 
-  // for some reason only this works with function for ticks
-  function getDefaultReferencePoint2() {
-    if ($store.ref) {
-      const found = data.find((d) => d[aggregationLevel] === $store.ref);
-      if (found) {
-        console.log(found);
-        return found;
-      }
-    }
-    return Object.values(getDefaultReferencePoint()['percentiles']);
-  }
-
-  // so that it doesn't jump say if we use defaultreferencepoint instead
-  const tryThis = (arr) => {
-    let something = _.round(arr.length / 5)
-    return [arr[0], arr[something], arr[2*something], arr[3*something], arr[arr.length - 1]]
-  }
+  const simplifyYDomain = (arr) => {
+    let partition = _.round(arr.length / 5);
+    return [
+      arr[0],
+      arr[partition],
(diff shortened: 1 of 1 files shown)


src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -143,6 +143,28 @@
...
+
+  const getUniqueArray = (arr) => {
+    let a = arr.concat();
+    for (let i = 0; i < a.length; ++i) {
+      for (let j = i + 1; j < a.length; ++j) {
+        if (a[i] === a[j]) a.splice(j--, 1);
+      }
+    }
+    return a;
+  };
```
robhudson, line 173: This looks like something that could go in `src/utils/` and a test written for it. Or, I'm curious if https://lodash.com/docs/3.10.1#uniq accomplishes the same thing?

robhudson (commented):
Definitely on the right track. This looks like it works really well and what we were hoping for. :+1:

Iinh:
If y-axis has less than 5 data points, the numerical range disappears. I still need to figure out where this condition is set to disable it.

src/components/explore/AggregationsOverTimeGraph.svelte
Iinh: I think it's the same thing. Thanks for the tip!

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -143,6 +135,64 @@
...
   }
+
+  function getDefaultReferencePoint() {
+    if ($store.ref) {
+      const found = data.find((d) => d[aggregationLevel] === $store.ref);
+      if (found) return found;
+    }
+    return data[data.length - 1];
+  }
+
```
robhudson: There's a lot happening in these 3 functions below. Adding a generalized comment above each one, and some comments within them to explain what is going on would be really helpful for future maintainers and ourselves.

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -143,6 +143,100 @@
...
     $showContextMenu = true; // eslint-disable-line no-unused-vars
   }
+
+  const getYTicks = (ranges) => {
+    if (
+      data[0].metric_type === 'histogram-linear' ||
+      data[0].metric_type === 'histogram-exponential'
+    ) {
+      // when the range is too small we need to custom set it
+      if (ranges[ranges.length - 1] < 5) return [0, 1, 2, 3, 4, 5];
```
robhudson: lodash has a `_.range(6)` that can return this, but not sure if that's better or not. If it were longer I'd say use lodash, but for just 6 numbers this doesn't seem bad.

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -143,6 +143,96 @@
...
+      yScaleType !== 'scalePoint' &&
+      buckets.length
+    ) {
+      if ($store.proportionMetricType === 'proportions') {
+        yDomainValues = yDomainValues.filter((a) => a < 1);
+        return [yDomainValues[0], yDomainValues[yDomainValues.length - 1]];
+      }
+      if ($store.proportionMetricType === 'counts')
+        return [yDomainValues[0], yDomainValues[yDomainValues.length - 1]];
+    }
```
robhudson, line 228: It seems like these are doing about the same thing and could be combined. Same with linear and histogram-linear?

Iinh: you're right. thanks for catching this

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -143,6 +143,95 @@
...
+      data[0].metric_type === 'histogram-linear' ||
+      yScaleType === 'scalePoint'
+    ) {
+      // eslint-disable-next-line no-restricted-syntax, guard-for-in
+      for (const p in visiblePercentiles) {
+        yData = [
+          ...yData,
+          ...data.map((arr) => arr.percentiles[visiblePercentiles[p]]),
+        ];
+      }
```
robhudson: This could be rewritten using `forEach` to satisfy eslint:
```js
      visiblePercentiles.forEach((p) => {
        yData = yData.concat(
          [...data.map((arr) => arr.percentiles[p])]
        );
      });
```

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -143,6 +143,95 @@
...
+  const getYDomain = (visiblePercentiles, buckets) => {
+    let yData = [];
+    let yDomainValues = [];
+
+    // linear and log graphs
+    if (
+      data[0].metric_type === 'histogram-linear' ||
+      yScaleType === 'scalePoint'
+    ) {
+      // eslint-disable-next-line no-restricted-syntax, guard-for-in
```
robhudson: Here's another one. Did this get re-introduced somehow?

robhudson (approved):
This looks good to me. Thanks for letting me comment randomly while I understood the changes. Let's get it on dev for others to play with!

PR #1416, opened 2021-06-08
Upgrade OpenResty + Support Redis

robotblake:
* Upgrade OpenResty to 1.19.3.1
* Upgrade lua-resty-openidc to 1.7.4
* Add Redis config options to the template
* Explicitly close the session after authenticating, this fixes timeouts
  that occur if the session store uses locking


dockerfiles/proxy/access_by_openidc.lua
```diff
@@ -24,7 +24,8 @@ if ngx.var.oidc_hd then
 end
 
 -- call authenticate for OpenID Connect user authentication
-local res, err = require("resty.openidc").authenticate(opts)
+local res, err, target, session = require("resty.openidc").authenticate(opts)
+session:close()
```
robhudson: Can we add a comment here for future people to know why this is necessary?

robhudson (changes requested):
r+wc

PR #1437, opened 2021-06-22
Fix rescaling y-axis edge cases

Iinh:
This PR fixes 3 things that we noticed after testing dev:

1. Use transformed percentile data for Android metrics. For example, the graph for [this metric](https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/startup_timeline_framework_primary/explore?ping_type=*) doesn't display any data because we were using data from the regular percentile source (like we did for telemetry probes), instead of sourcing from `transformedPercentiles`, which is for Android metrics.

2. Add more metrics to linear graph group, such as counter and labeled_counter. This will let us rescale y-axis for these metrics too, since currently they do not on dev (see examples for [counter](https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/events_total_uri_count/explore?ping_type=*) and [labeled_counter](https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/logins_store_unlock_error_count/explore?ping_type=*&timeHorizon=ALL))

3. @rafrombrc pointed out that we should keep the y domain range to match the violin plot at first, and only rescale it when a percentile is toggled. I added a condition to fix this.

### DIFF
--- a/src/components/explore/AggregationsOverTimeGraph.svelte
+++ b/src/components/explore/AggregationsOverTimeGraph.svelte
@@ -144,10 +144,17 @@
     $showContextMenu = true; // eslint-disable-line no-unused-vars
   }
 
+  const linearMetrics = [
+    'histogram-linear',
+    'quantity',
+    'counter',
+    'labeled_counter',
+  ];
+
   const getYTicks = (ranges) => {
+    // exponential and linear graphs
     if (
-      data[0].metric_type === 'histogram-linear' ||
-      data[0].metric_type === 'quantity' ||
+      linearMetrics.includes(data[0].metric_type) ||
       yScaleType === 'scalePoint'
     ) {
       // when the range is too small we need to custom set it
@@ -155,6 +162,7 @@
       if (yScaleType === 'scalePoint' && ranges.length < 5) return yValues;
       return undefined;
     }
+    // categorical graphs
     if (
       $store.activeBuckets.length &&
       $store.proportionMetricType === 'proportions'
@@ -178,13 +186,23 @@
 
     // get percentile data of linear and log graphs
     if (
-      data[0].metric_type === 'histogram-linear' ||
-      data[0].metric_type === 'quantity' ||
+      linearMetrics.includes(data[0].metric_type) ||
       yScaleType === 'scalePoint'
     ) {
+      // do not change yDomain when all percentiles are selected
+      // (we want the graph to match the violin plot initially)
+      if ($store.visiblePercentiles.length === 5) return yDomain;
       visiblePercentiles.forEach((p) => {
         yData = yData.concat([...data.map((arr) => arr.percentiles[p])]);
       });
+      // use transformed data for Android metrics
+      if (data[0].transformedPercentiles) {
+        visiblePercentiles.forEach((p) => {
+          yData = yData.concat([
+            ...data.map((arr) => arr.transformedPercentiles[p]),
+          ]);
+        });
+      }
     }
     // get proportion and count data of categorical graphs
     if (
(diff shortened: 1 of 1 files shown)


emtwo (commented):
Code looks good to me! Would you mind putting some screenshots of the changes you made for reference. Helpful especially since I'm not super familiar with the codebase. Thank you! :)

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -144,17 +144,25 @@
     $showContextMenu = true; // eslint-disable-line no-unused-vars
   }
 
+  const linearMetrics = [
+    'histogram-linear',
+    'quantity',
+    'counter',
+    'labeled_counter',
+  ];
```
Iinh, line 152: These metrics have linear graph type, but initially I only had `histogram-linear` and `quantity` on this list, so other linear graphs such as `counter` and `labeled_counter` did not scale (see [counter](https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/events_total_uri_count/explore?ping_type=*) and [labeled_counter](https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/logins_store_unlock_error_count/explore?ping_type=*&timeHorizon=ALL))

After adding them to the list, we can rescale y-axis for `counter` and `labeled_counter` now. 

Working example of a `counter` metric:

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -178,20 +186,32 @@
...
     // get percentile data of linear and log graphs
     if (
-      data[0].metric_type === 'histogram-linear' ||
-      data[0].metric_type === 'quantity' ||
+      linearMetrics.includes(data[0].metric_type) ||
       yScaleType === 'scalePoint'
     ) {
+      // do not change yDomain when all percentiles are selected
+      // (we want the graph to match the violin plot initially)
+      if ($store.visiblePercentiles.length === 5) return yDomain;
```
Iinh, line 194: We were not able to rescale the y axis because we used to have the same set of data for yDomain regardless of toggled percentiles. In my last change, I calculated the yDomain range to only include the highest and smallest values available of the shown percentile lines, so that this array will dynamically change if a percentile is toggled. However this means that the range is more tightly scoped than before, and you can see the difference in initial behavior here: [dev](https://dev.glam.nonprod.dataops.mozgcp.net/firefox/probe/time_to_non_blank_paint_ms/explore?) vs [stage](https://stage.glam.nonprod.dataops.mozgcp.net/firefox/probe/time_to_non_blank_paint_ms/explore?).

@rafrombrc wanted the initial range of the graph to match the violin plot graph, so I added this condition here to preserve the yDomain range if all percentile lines are visible (i.e no percentiles toggled yet)

Below you can see that the yDomain range matches the violin plot initially.

match

src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -178,20 +186,32 @@
...
         yData = yData.concat([...data.map((arr) => arr.percentiles[p])]);
       });
+      // use transformed data for Android metrics
+      if (data[0].transformedPercentiles) {
+        visiblePercentiles.forEach((p) => {
+          yData = yData.concat([
+            ...data.map((arr) => arr.transformedPercentiles[p]),
+          ]);
+        });
+      }
```
Iinh, line 205: Currently on dev [this metric](https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/startup_timeline_framework_primary/explore?ping_type=*) doesn't have any data. After changing data source to `$store.transformedPercentiles` (which I believe is the main source of data for Glean metrics, we use `$store.percentiles` for desktop probes), it does now.

Iinh (commented):
@emtwo Of course! I included some screenshots below along with my thought process behind this change.

emtwo (approved):
This looks great! Thanks so much for the screen captures, @Iinh!

PR #1466, opened 2021-07-12
Increase graphic container width to show bigger numbers

Iinh:
This PR increases the width of the DataGraphic container in order to show bigger values. See original PR in `graph-paper`: https://github.com/graph-paper-org/graph-paper/pull/365 for the before and after.

As discussed with @robhudson, I copied the DataGraphic component from graph-paper to make the width change directly in GLAM. This is the start of an effort to pull all graph-paper components back into glam.


src/components/explore/AggregationComparisonGraph.svelte
```diff
@@ -1,9 +1,8 @@
 <script>
   import { Axis } from '@graph-paper/guides';
 
-  import { DataGraphic } from '@graph-paper/datagraphic';
-
   import { tooltip as tooltipAction } from '@graph-paper/core/actions';
+  import DataGraphic from '../../graph-paper/datagraphic/DataGraphic.svelte';
```
robhudson: I think we can drop the `graph-paper/` folder here. Or do you feel there's a reason for it?

package.json
```diff
@@ -57,6 +57,7 @@
     "svelte-loader": "3.1.2"
   },
   "dependencies": {
+    "d3-array": "^3.0.1",
```
robhudson: Let's drop the `^` to match the rest.

src/graph-paper/datagraphic/DataGraphic.svelte
```diff
@@ -0,0 +1,558 @@
...
+    justify-content: center;
+  }
+</style>
+
+<div
+  bind:this={container}
+  class="data-graphic-container"
+  style="width: {$graphicWidth}px; height: {$graphicHeight}px;">
+  <svg
+    style="width: {$graphicWidth + 90}px; height: {$graphicHeight}px;"
```
robhudson: Just noting that here is the change from the original DataGraphic.svelte.

src/components/explore/AggregationComparisonGraph.svelte
Iinh: I intend to use this `graph-paper/` folder for all the components that we're going to pull. Wdyt?

robhudson: Okay. I think "graph-paper" doesn't really have meaning inside glam. And it's an empty folder. I think the graph-paper components should all live with the other components inside `src/components` and subfolders there.

Iinh: Makes sense. I'll move DataGraphics to `src/components/` then

robhudson (approved):

PR #1476, opened 2021-07-19
Properly reset default state on homepage

Iinh:
Since the default state is initialized with information from URL query params, we need to properly reset store when user clicks on home page so that state of one probe doesn't get transferred to another.

### DIFF
--- a/src/state/store.js
+++ b/src/state/store.js
@@ -71,8 +71,12 @@ function getDefaultState(
 
 export const store = createStore(getDefaultState({ basedOnQueryParams: true }));
 
-store.reset = () => {
+store.reset = (home = false) => {
   store.reinitialize();
+  if (home) {
+    // if user clicks on home page, reset state to not include query params
+    store.setState(getDefaultState());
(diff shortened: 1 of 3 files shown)


src/state/store.js
```diff
@@ -71,8 +71,12 @@ function getDefaultState(
 
 export const store = createStore(getDefaultState({ basedOnQueryParams: true }));
 
-store.reset = () => {
+store.reset = (home = false) => {
   store.reinitialize();
+  if (home) {
+    // if user clicks on home page, reset state to not include query params
+    store.setState(getDefaultState());
+  }
```
robhudson, line 80: It looks like maybe the `setState` call overrides and doesn't need the `reinitialize` call above it if it's `home`. If that's true, can we skip that call and only `reinitialize` in an `else` clause?

Iinh: makes sense! i just updated the logic here in a new commit

robhudson (approved):
Looks good to me

PR #1486, opened 2021-07-22
Update deployment docs

Iinh:


docs/deployment.md
```diff
@@ -32,7 +34,12 @@ The steps for tagging a release for the staging server are as follows:
 3. Tag a release locally in git via: `git tag <tag>`
 4. Push the tag to github via: `git push origin --tags`
 
-## Logs
+## How to view the import logs in Google Cloud Platform console
 
-The results of the deploy will be logged to the **#datatools-deploys** channel
-on Slack.
```
robhudson, line 38: This is still true and doesn't need to be deleted.

docs/deployment.md
```diff
@@ -32,7 +34,12 @@ The steps for tagging a release for the staging server are as follows:
...
+1. Log into the GCP console web page using authorized credentials
+2. Make sure you're in the "glam-prod" project
+3. Drill down into the "Kubernetes Engine -> Workloads"
+4. Look for the one that says name="glam-prod-glam-app-1" type="Cron job" and
+   click the name
+5. Select the "Logs" tab along the top. From there you can see all the recent
+   logging from the imports which are run via cron.
```
robhudson, line 45: :+1:

docs/deployment.md
Iinh: I think since @haroldwoo took over we don't log the deploy status to #datatools-deploys anymore, at least from what i've seen so far in that channel. Since we're using jira to communicate with SRE now, the last 2 succesful deploys were not logged to slack, but rather marked DONE in the JIRA ticket.

robhudson: Okay, thanks. That's new to me. In that case, this is good to go. :)

robhudson (approved):

PR #1531, opened 2021-08-23
Persist aggregation type and key in URL

### LINKED ISSUE #1182: Aggregation dropdown value should be encoded in the URL
This would be helpful when sharing links so the aggregation choice is shared as well

Iinh:
Fixes #1182. Selected aggregation types and keys are now persisted in the URL.

### DIFF
--- a/src/components/explore/QuantileExplorerView.svelte
+++ b/src/components/explore/QuantileExplorerView.svelte
@@ -17,6 +17,7 @@
     gatherProbeKeys,
     gatherAggregationTypes,
   } from '../../utils/probe-utils';
+  import { store } from '../../state/store';
 
   const dispatch = createEventDispatcher();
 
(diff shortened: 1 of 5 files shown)


src/components/explore/QuantileExplorerView.svelte
```diff
@@ -32,10 +33,10 @@
   let aggregationTypes = gatherAggregationTypes(data);
   let probeKeys = gatherProbeKeys(data);
 
-  let currentKey = probeKeys[0];
+  let currentKey = $store.aggKey || probeKeys[0];
   let currentAggregation = aggregationTypes.includes('summed_histogram')
     ? 'summed_histogram'
-    : 'avg';
+    : $store.aggType;
```
robhudson, line 39: If there isn't one stored does it still default to `'avg'`?

Iinh: It should still default to 'avg'. See: https://github.com/mozilla/glam/pull/1531/files#diff-352b8684efbebfa2b2a01b4af9892eea7fb1c2f2f9cd6c4acf803c7fa8974c32R40

robhudson (approved):

PR #1533, opened 2021-08-24
Filter boolean data before passing down to component

Iinh:
After merging #1525 to dev yesterday I noticed that it didn't completely fix the problem for _all_ boolean metrics (for example [this](https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/metrics_has_open_tabs/explore?activeBuckets=%5B%22always%22%2C%22never%22%2C%22sometimes%22%5D&aggType=avg) works but [this](https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/metrics_has_desktop_bookmarks/explore?aggType=avg) doesn't). After digging more I realize that it's best to filter the boolean data before we pass it down to our components, i.e. in our fenix store, so that the data we get is consistent across all metrics. 

With this change I was able to see locally that graphs for boolean metrics are working again.

### DIFF
--- a/src/config/fenix.js
+++ b/src/config/fenix.js
@@ -131,21 +131,16 @@ export default {
     // the frontend. It will always run, even against cached data, as a way of
     // resetting the necessary state.
     const viewType = this.probeView[data[0].metric_type];
-
-    const isCategoricalTypeProbe = viewType === 'categorical';
-
     let etc = {};
 
     // filter out true/false aggregate results in boolean metrics. See: https://github.com/mozilla/glam/pull/1525#discussion_r694135079
     if (data[0].metric_type === 'boolean') {
+      // eslint-disable-next-line no-param-reassign
       data = data.filter((di) => di.client_agg_type === '');
     }
 
-    if (isCategoricalTypeProbe) {
+    if (viewType === 'categorical') {
       etc = extractBucketMetadata(data);
-    }
-
-    if (isCategoricalTypeProbe) {
       appStore.setField('activeBuckets', etc.initialBuckets);
     }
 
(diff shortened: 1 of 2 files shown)


src/config/fenix.js
```diff
@@ -135,6 +135,12 @@ export default {
     const isCategoricalTypeProbe = viewType === 'categorical';
 
     let etc = {};
+
+    // filter out true/false aggregate results in boolean metrics. See: https://github.com/mozilla/glam/pull/1525#discussion_r694135079
+    if (data[0].metric_type === 'boolean') {
```
robhudson, line 137: They `metric_type` is being accessed above and stored as `viewType`. You can use that here.

src/config/fenix.js
```diff
@@ -135,6 +135,12 @@ export default {
...
+      data = data.filter((di) => di.client_agg_type === '');
+    }
+
```
robhudson, line 141: Down below there's 2, `if (isCategoricalTypeProbe)`. Can we collapse that into a single one? And if that's the only place `isCategoricalTypeProbe` is used I'd say change it to `if (viewType === 'categorical')` to match your new code. No sense having a variable just for one conditional.

src/config/fenix.js
Iinh: I don't think they're the same: in this example for boolean metrics `data[0].metric_type` is `boolean` while `viewType` is `categorical`.

robhudson: Ah, my eyes skipped over the `this.probeView` part of it. Apologies.

robhudson (approved):

PR #1558, opened 2021-09-07
Show absolute difference when comparing percentage data points (fixes #1136)

### LINKED ISSUE #1136: Need absolute differences when comparing content_frame_time_reason data points
When comparing [two datapoints for the `content_frame_time_reason` probe](https://glam.telemetry.mozilla.org/firefox/probe/content_frame_time_reason/explore?activeBuckets=%5B%22OnTime%22%2C%22SlowComposite%22%2C%22MissedCompositeLong%22%2C%22MissedCompositeMid%22%2C%22MissedComposite%22%2C%22MissedCompositeLow%22%2C%22NoVsyncNoId%22%2C%22NoVsync%22%5D&process=parent&ref=20210115035053), the "DIFF" column in the summary table is showing useless values.

Screen Shot 2021-01-26 at 12 37 40 PM

Basically, we're comparing two pie charts here. The "OnTime" percentage has gone down from 82.26% to 78.76%, and I'm trying to find out where those 3.5% percentage points went.

So those 3.5% must somehow be absorbed by the other rows.
A table with absolute differences would be much more useful:

| Category | Hov | Ref | Diff |
|---|---|---|---|
| OnTime | 82.26% | 78.76% | -3.5% |
| SlowComposite | 5.36% | 3.57% | +1.79% |
| MissedCompositeLong | ... | ... | ... |
| ... | ... | ... | ... |

(The sum of the Diff column should be 0%, because each column individually sums up to 100%.)

Iinh:
Currently we're showing the relative change (in percentage) in the "DIFF." column in the summary table. This works in the case of exponential and linear probes with quantile data points. However, categorical probes have percentage data, therefore showing relative change isn't as helpful. This pain point is reflected in #1136. To improve this, for categorical probe types, we should switch to calculating actual differences for the DIFF. colum (Δ = ref. - hov.) instead of relative change (ref. - hov. / ref.)

### DIFF
--- a/src/components/explore/ComparisonSummary.svelte
+++ b/src/components/explore/ComparisonSummary.svelte
@@ -18,9 +18,10 @@
   export let showLeft = true;
   export let showRight = true;
   export let showDiff = true;
+  export let viewType;
 
   function percentChange(l, r) {
-    return (r - l) / l;
+    return viewType === 'proportion' ? r - l : (r - l) / l;
   }
 
   let displayValues = [];
--- a/src/components/explore/ProbeExplorer.svelte
+++ b/src/components/explore/ProbeExplorer.svelte
@@ -322,7 +322,8 @@
     valueFormatter={summaryNumberFormatter}
     keyFormatter={comparisonKeyFormatter}
     showLeft={data.length > 1}
-    showDiff={data.length > 1} />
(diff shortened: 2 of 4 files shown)


robhudson (approved):
Looks good :+1:

PR #1561, opened 2021-09-07
Add docs describing the imports and db schema

robhudson:


robhudson:
@Iinh @alekhyamoz Let me know if there's anything unclear or if anything could be expanded upon or I missed anything wrt imports here. I'll leave this PR open until this is complete. Thanks!

docs/data-imports.md
```diff
@@ -0,0 +1,52 @@
...
+based product and will be the example for future products in GLAM going forward.
+
+## Schema
+
+The Firefox for Desktop (aka fenix) product has 2 primary tables assocated with
+its data. The aggregation table and the materialized view on top of that data.
+
+The aggregation table exist so imports can be as fast as possible. It has
+minimal indexes and constraints so data can import quickly. The only index is
+the primary key. The only constraint is a unique constraint comprised of the
```
Iinh: > The only index is the primary key.

So the only index for the aggregation table is the auto-generated id right?

docs/data-imports.md
```diff
@@ -0,0 +1,52 @@
...
+dimensions of the data (e.g. channel, build ID, metric, etc.) so that we can
+assure we have only 1 row across these dimensions. This unique constraint is
+also used in the "upsert" so that if new data comes in we insert it, or if a
+matching record exists we update the data (counts, histograms, and percentiles).
+
+The materialized view exists so the API queries are as fast as possible. It is a
+direct copy of the aggregation table but with added indexes, roughly to match
+the top level selectors in GLAM so the SQL `WHERE` clauses are fast. Since the
```
Iinh: > It is a direct copy of the aggregation table but with added indexes, roughly to match the top level selectors in GLAM so the SQL `WHERE` clauses are fast.

Where in the code is this part `it is a direct copy of the aggregation table but with added indexes`?

docs/data-imports.md
```diff
@@ -0,0 +1,52 @@
...
+
+- **Create a list of files** to import by pattern matching all files in the GCS
+  bucket, matching for Fenix and the app ID
+- For each file:
+  - we **create a temporary table** and **load the CSV data** into this
+    temporary table using the Postgresql `COPY` command,
+  - then we **perform an "upsert"** to quickly move the data into the main
+    aggregation table
+- Once all files have been imported, we **refresh the materialized view** so the
+  new data is shared via the API
```
Iinh: If it's not too much work, could you reference the code for each step here?

docs/data-imports.md
robhudson: Right. I can try to clarify that some there.

docs/data-imports.md
robhudson: It is in the migrations that created the tables. I could add a link to it here, but if those files ever change this would need to be updated.

docs/data-imports.md
robhudson: Yep, I can add that.

robhudson: I ended up not linking to each part of the file directly, but tried to align the description better with the code so it's easier to follow along. My fear is the code will change out from under the docs and make the docs more confusing.

PR #1586, opened 2021-09-20
Add backfilled ADR

robhudson:


docs/adr/0003-file-based-data.md
```diff
@@ -0,0 +1,88 @@
...
+    ).fetchall()
+    open(os.path.join(outdir, 'data.json'), 'w').write(json.dumps(aggregates))
+```
+
+#### Pros/cons:
+
+- Pro: This could be a simpler local dev set up.
+- Pro: This could improve the slow import times due to the materialized view.
+- Pro: This could make adding new products to glam a bit simpler -- no need to
+  create new Django models.
```
wlach, line 73: Is this actually true? I don't see why we would *need* seperate models (tables) per application with the current approach (although maybe it makes more sense?).

docs/adr/0003-file-based-data.md
```diff
@@ -0,0 +1,88 @@
+# Use files for metrics data
```
wlach, line 1:
```suggestion
# Using files instead of a database to store aggregated metrics data
```

docs/adr/0003-file-based-data.md
```diff
@@ -0,0 +1,88 @@
...
+
+- Status: rejected
+- Date: 2020 (this ADR was backfilled for historical purposes)
+
+## Context and Problem Statement
+
+The GLAM data consists mainly of a histogram, percentiles, and counts for each
+set of dimensions. This can be stored in a file whose path and name can
+represent the dimensions and whose content can contain the actual data.
```
wlach, line 10: Worth mentioning the current postgres-based architecture. I recommend at least linking to your google doc:

https://docs.google.com/document/d/1JN4XQdG-8S5wbRUWAFp0XpDfqNu9ti_rdmy8OZXuEyE/edit

(that would in itself be a good ADR, though I understand if you don't have time to reformat it)

docs/adr/0003-file-based-data.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Decision Drivers
+
+- The data should be efficient to look up and read for the front-end
+- The data should be efficient to write for the ETL
+- The data should be cost efficient to write and maintain
+
+## Considered Options
+
+- Move aggregation data, the output from the ETL, into a Postgresql database
```
wlach, line 20:
```suggestion
- Keep aggregated data, the output from the ETL, in a Postgresql database
```

docs/adr/0003-file-based-data.md
```diff
@@ -0,0 +1,88 @@
...
+- Move aggregation data into another file format more usable for the GLAM
+  front-end
+
+## Decision Outcome
+
+Moving the data into Postgresql was chosen, as outlined in the pros/cons below.
```
wlach, line 26:
```suggestion
Keeping the data in Postgresql was chosen, as outlined in the pros/cons below.
```

docs/adr/0003-file-based-data.md
```diff
@@ -0,0 +1,88 @@
...
+- Pro: This could make adding new products to glam a bit simpler -- no need to
+  create new Django models.
+- Con: Since data outside the range of the ETL needs to be kept, this can't be a
+  simple write operation -- the fresh data needs to be added or updated while
+  keeping older data and the resulting data needs to be written out. This
+  complicates the file based architecture.
+
+### Database aggregate storage
+
+With this approach we would write the data to an unindexed Postgresql table to
```
wlach, line 81: This is written as if it's an option we're considering, rather than the current state of things, which is a bit confusing.

```suggestion
This is the current approach. We write the data to an unindexed Postgresql table to
```

wlach (approved):
This looks good to me, I have a few minor suggestions.

It would be IMO good for @alekhyamoz to weigh in here, since this is an area of GLAM she'll be interacting with in the future.

alekhyamoz (approved):

PR #1599, opened 2021-09-28
Add histogram distribution view to the table view (fixes #1560)

### LINKED ISSUE #1560: Add histogram distribution view to the table view
The table view currently only shows the percentile values for each build, it would be useful to be able to toggle btn those and the full histogram distribution values (that we already have in the GLAM CloudSQL tables) so folks can see that as well.

Iinh:
The goal of this PR is to surface distribution data in the table view for metrics that currently only show percentile data. The metric types are:

- desktop: exponential, linear, keyed-scalar, scalar
- fenix: quantity, counter, labeled_counter

Other metric types already have distribution data in table view. For example: [categorical](https://glam.telemetry.mozilla.org/firefox/probe/dom_script_loading_source/table?activeBuckets=%5B%22Inline%22%2C%22Source%22%2C%22AltData%22%2C%22SourceFallback%22%5D), [enumerated](https://glam.telemetry.mozilla.org/firefox/probe/pwmgr_prompt_remember_action/table?activeBuckets=%5B%220%22%2C%222%22%2C%221%22%2C%223%22%2C%224%22%5D&process=parent), [boolean](https://glam.telemetry.mozilla.org/fenix/probe/preferences_open_links_in_app_enabled/table?activeBuckets=%5B%22never%22%2C%22always%22%2C%22sometimes%22%5D&aggType=avg&ping_type=*), etc. 

Before:

CleanShot 2021-09-30 at 14 29 56@2x

After:

### DIFF
--- a/src/components/table/TableView.svelte
+++ b/src/components/table/TableView.svelte
@@ -79,7 +79,7 @@
       {currentPage} />
   </div>
 
-  {#if metricsWithPercentileData.includes(data[0].metric_type)}
+  {#if METRICS_WITH_PERCENTILE_DATA.has(data[0].metric_type)}
     <div style="display: flex; justify-content: flex-end; padding: 1em;">
       <ButtonGroup>
         <Button
--- a/src/components/table/TableView.svelte
+++ b/src/components/table/TableView.svelte
@@ -1,4 +1,6 @@
 <script>
+  import { Button, ButtonGroup } from '@graph-paper/button';
+
   import DataTable from './DataTable.svelte';
   import Row from './Row.svelte';
   import Cell from './Cell.svelte';
@@ -12,6 +14,7 @@
     formatPercentDecimal,
     ymd,
     timecode,
+    formatPercent,
   } from '../../utils/formatters';
 
   import { backwards } from '../../utils/iterables';
@@ -27,10 +30,22 @@
   export let pageSize = 10;
   export let bucketTypeLabel = 'Categories';
 
+  let showHistogramData = false;
+
   let totalPages = 0;
   let currentPage = 0;
   $: if (data) currentPage = 0;
   $: totalPages = Math.ceil(data.length / pageSize);
+
+  const metricsWithPercentileData = [
+    'histogram-exponential',
+    'histogram-linear',
+    'keyed-scalar',
+    'scalar',
+    'quantity',
+    'counter',
+    'labeled_counter',
+  ];
 
   let largestAudience;
   $: largestAudience = Math.max(...data.map((d) => d.audienceSize));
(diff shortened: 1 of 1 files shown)


src/components/table/TableView.svelte
```diff
@@ -100,20 +134,36 @@
...
+              topBorder={true}
+              bottomBorderThickness="2px">
+              <span
+                class="percentile-label-block"
+                style="background-color: black" />
+              <span class="bucket">{bucket.bin}</span>
+            </Cell>
+          {/each}
+        {:else}
+          {#each visibleBuckets as bucket}
```
wlach:
```suggestion
        {:else}
          <!-- show percentile data -->
          {#each visibleBuckets as bucket}
```

src/components/table/TableView.svelte
```diff
@@ -27,11 +30,23 @@
...
+  const metricsWithPercentileData = [
+    'histogram-exponential',
+    'histogram-linear',
+    'keyed-scalar',
+    'scalar',
+    'quantity',
+    'counter',
+    'labeled_counter',
+  ];
+
```
wlach: I prefer to uppercase constant definitions to make it more clear what they are. Also this is probably better expressed as a [set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).

(at some point we might want to shove this type of thing off into a constants file, but maybe doesn't need to happen here)

```suggestion
  const METRICS_WITH_PERCENTILE_DATA = new Set([
    'histogram-exponential',
    'histogram-linear',
    'keyed-scalar',
    'scalar',
    'quantity',
    'counter',
    'labeled_counter',
  ]);

```

src/components/table/TableView.svelte
```diff
@@ -63,15 +78,34 @@
       {totalPages}
       {currentPage} />
   </div>
+
+  {#if metricsWithPercentileData.includes(data[0].metric_type)}
```
wlach: (if you implement the above)

```suggestion
  {#if METRICS_WITH_PERCENTILE_DATA.has(data[0].metric_type)}
```

wlach (approved):
Looks great! We're doing this work on behalf of the performance team, correct? Might be worth getting their feedback on this UI before landing, to reduce back/forth (too bad we don't have deploy previews for GLAM)

Iinh:
> Looks great! We're doing this work on behalf of the performance team, correct? Might be worth getting their feedback on this UI before landing, to reduce back/forth

Thanks @wlach for the review! For this change specifically, we already ran it by the performance team. I'm planning on reaching out to them to give an update and ask for a review once a couple more features have landed.

PR #1617, opened 2021-10-05
Add link to comparison view in Redash from context menu (fixes #1577)

### LINKED ISSUE #1577: Add link to comparison view redash query from context menu
Bas's comparison view query is super useful. Ultimately, it'd be nice to have this in GLAM itself, but for now we can link out to the query by constructing a URL that pre-populates the parameters for the parameterized redash query.

Iinh:
Fixes #1577.

Feedback from the performance team: they often use GLAM as a SQL generator to run custom queries such as [#81693](https://sql.telemetry.mozilla.org/queries/81693/source?p_Build%201=2021-02-22T14%3A26%3A01&p_Build%202=2021-08-04T19%3A32%3A34&p_Days%20to%20Query=7&p_OS=Windows&p_Probe=payload.processes.gpu.histograms.keypress_present_latency&p_Start%20Date=2021-02-28&p_Start%20Date%202=2021-08-12#202508) to view stacked histogram comparison and CDF comparison.

Using Redash [query parameters](https://sql.telemetry.mozilla.org/queries/81693/source?p_Build%201=2021-02-22T14%3A26%3A01&p_Build%202=2021-08-04T19%3A32%3A34&p_Days%20to%20Query=7&p_OS=Windows&p_Probe=payload.processes.gpu.histograms.keypress_present_latency&p_Start%20Date=2021-02-28&p_Start%20Date%202=2021-08-12#202508), we can prepopulate the parameters in the URL and add a link to it from the context menu. 

Currently this feature is only available for Firefox probes that are histogram.

I changed Bas' query a little bit, namely: 

- 2 new params:
   - `table`: `telemetry.main_nightly` or `telemetry.main_1pct` to improve run time (originally it was `telemetry.main` which contains a lot of data and takes a while for each query to finish)
   - `channel`: nightly/beta/release

- the way we get build_is different. In Bas' query, the build_is is retrieved from dropdown values that are a result of another query. Here we directly get the values from the hovered and clicked positions on the graph. 

Bas':

CleanShot 2021-10-18 at 11 13 08@2x

Our query:

CleanShot 2021-10-18 at 11 13 42@2x

UI change:

CleanShot 2021-10-18 at 11 15 19@2x


src/components/ChartContextMenu.svelte
```diff
@@ -154,6 +183,18 @@
           </a>
         </div>
       </div>
+      {#if getComparisonViewinSTMO()}
```
emtwo: nit: `getComparisonViewinSTMO()` is being called 3 times but does it need to be? Perhaps we can make the if statement here check `$store.product === 'firefox'` and then the `getComparisonViewinSTMO()` function doesn't need to do that check and can be called just once below.

src/components/ChartContextMenu.svelte
```diff
@@ -154,6 +183,18 @@
...
+        <div class="option">
+          <div class="option-icon">
+            <a href={getComparisonViewinSTMO()}>
+              <Graphs />
+            </a>
+          </div>
+          <div class="option-link">
+            <a href={getComparisonViewinSTMO()}>View Comparison in STMO</a>
```
emtwo: Actually, `getComparisonViewinSTMO()` isn't passed any params. Can we turn it into a variable instead of a function so that the url string is only generated one time and used in both the `href` references above?

src/components/ChartContextMenu.svelte
```diff
@@ -48,6 +49,34 @@
...
+  const table =
+    $store.productDimensions.channel === 'nightly'
+      ? 'main_nightly'
+      : 'main_1pct';
+
+  const getComparisonViewinSTMO = () =>
+    $store.product === 'firefox'
+      ? 'https://sql.telemetry.mozilla.org/queries/82247/source?' +
+        `&p_Table=telemetry.${table}` +
+        `&p_Probe=${getTelemetryPath()}` +
```
emtwo: What happens when `getTelemetryPath()` returns `undefined`? I assume then for scalars the `href` below will be set to some invalid link and then you can click `View Comparison in STMO` but the link won't work. We may need some check to only show the button if the value is a histogram.

emtwo (commented):
This looks cool! Just left a couple of comments

src/components/ChartContextMenu.svelte
```diff
@@ -48,6 +49,34 @@
...
+  };
+
+  const table =
+    $store.productDimensions.channel === 'nightly'
+      ? 'main_nightly'
+      : 'main_1pct';
+
+  const getComparisonViewinSTMO = () =>
+    $store.product === 'firefox'
+      ? 'https://sql.telemetry.mozilla.org/queries/82247/source?' +
```
wlach: I think this would be slightly easier to read if you constructed it with [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

src/components/ChartContextMenu.svelte
wlach: Making this a [reactive declaration](https://svelte.dev/tutorial/reactive-declarations) seems like the right way forward to me.

wlach (changes requested):
This looks pretty good to me. A few small things aside from what @emtwo mentioned.

src/components/ChartContextMenu.svelte
```diff
@@ -48,6 +49,44 @@
...
+  const table =
+    $store.productDimensions.channel === 'nightly'
+      ? 'main_nightly'
+      : 'main_1pct';
+
+  const REDASH_PROBE_COMPARISON_URL =
+    'https://sql.telemetry.mozilla.org/queries/82247/source?';
+
+  const getComparisonViewinSTMO = (clicked, hovered) => {
+    let queryParams = new URLSearchParams();
```
wlach: According to the [URLSearchParams examples on MDN](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#examples), you should be able to create them all in one go. E.g.:

```suggestion
    const queryParams = new URLSearchParams({ p_Table: `telemetry.${table}`, p_Probe: getTelemetryPath(), ...});
```

wlach (approved):
r+wc

src/components/ChartContextMenu.svelte
Iinh: Ah yes, this is a nicer way to construct it. Thanks @wlach!

emtwo (approved):
lgtm!

PR #1619, opened 2021-10-06
Show 99 and 99.9 percentiles

Iinh:
Since we're getting new 99 and 99.9 percentiles for telemetry probes soon (after https://github.com/mozilla/bigquery-etl/pull/2412 is merged), this work is to get the front-end ready.

### DIFF
--- a/src/components/controls/PercentileSelectionControl.svelte
+++ b/src/components/controls/PercentileSelectionControl.svelte
@@ -3,6 +3,7 @@
   import BodyControl from './BodyControl.svelte';
 
   export let percentiles = [50];
+  export let product;
 
   function t(v) {
     // We can even return a string built using a template literal
@@ -40,6 +41,20 @@
       labelColor: cmp(95),
       tooltip: t(95),
       enabled: true,
+    },
+    {
+      label: '99%',
+      value: 99,
+      labelColor: cmp(99),
+      tooltip: t(99),
+      enabled: product === 'firefox',
+    },
+    {
+      label: '99.9%',
+      value: 99.9,
+      labelColor: cmp(99.9),
+      tooltip: t(99.9),
+      enabled: product === 'firefox',
     },
   ]}
   reverse={true}
--- a/src/components/controls/PercentileSelectionControl.svelte
+++ b/src/components/controls/PercentileSelectionControl.svelte
@@ -3,7 +3,6 @@
   import BodyControl from './BodyControl.svelte';
 
   export let percentiles = [50];
-  export let product;
 
   function t(v) {
     // We can even return a string built using a template literal
@@ -47,14 +46,14 @@
       value: 99,
       labelColor: cmp(99),
       tooltip: t(99),
-      enabled: product === 'firefox',
+      enabled: true,
     },
     {
       label: '99.9%',
       value: 99.9,
       labelColor: cmp(99.9),
       tooltip: t(99.9),
-      enabled: product === 'firefox',
+      enabled: true,
     },
   ]}
   reverse={true}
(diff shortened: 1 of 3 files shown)


Iinh:
Example:

By default `99` and `99.9` percentiles are off per @rafrombrc's suggestion.

CleanShot 2021-10-07 at 13 54 58@2x

When they are toggled on:

CleanShot 2021-10-07 at 13 55 14@2x

Iinh:
@alekhyamoz just confirmed that she's working on getting 99 and 99.9 data for fenix metrics in https://github.com/mozilla/bigquery-etl/pull/2412 too, so I updated the code to also add the new toggles to fenix in the latest commit.

wlach (changes requested):
Hmm, I can't seem to get this working. I'm seeing these errors in the console after a loading a test page (e.g. http://localhost:8000/firefox/probe/webrtc_audio_quality_outbound_rtt/explore?visiblePercentiles=%5B99%2C95%2C75%2C50%2C25%2C5%5D):

image

A couple of random points:

1. Based on the screenshot, I think you should adjust the scale of the graph depending on if 99/99.9 are selected. As it is the scale is distorted even if they are deselected.
2. I think we ought to add a help bubble by the "percentiles" to give a user an idea of what they mean (and also draw attention to the fact that they can be changed). This could be done as a follow-up, of course.

Iinh:
> Hmm, I can't seem to get this working. I'm seeing these errors in the console after a loading a test page

@wlach could you try importing data again (`./manage.py import_desktop_aggs <CHANNEL>` and `./manage.py import_glean_aggs <PRODUCT>`)? The screenshot was from data that I manually imported for testing purposes, but we should already have the new percentile data available in GCP, for both telemetry and fenix probes.

> Based on the screenshot, I think you should adjust the scale of the graph depending on if 99/99.9 are selected. As it is the scale is distorted even if they are deselected.

By default, upon landing on the page, the scale is set up so that it matches the violin plot range (this was @rafrombrc's request.) Since by 99/99.9 are deselected by default, the scale is kept as is, but selecting them should readjust the scale accordingly.

> I think we ought to add a help bubble by the "percentiles" to give a user an idea of what they mean (and also draw attention to the fact that they can be changed). This could be done as a follow-up, of course.

We do tooltip the percentiles. Do you think we should add more to it?

CleanShot 2021-10-08 at 14 06 59@2x

wlach:
> > Hmm, I can't seem to get this working. I'm seeing these errors in the console after a loading a test page
> 
> @wlach could you try importing data again (`./manage.py import_desktop_aggs <CHANNEL>` and `./manage.py import_glean_aggs <PRODUCT>`)? The screenshot was from data that I manually imported for testing purposes, but we should already have the new percentile data available in GCP, for both telemetry and fenix probes.

Ah got it, it works now! 🎉 

> > Based on the screenshot, I think you should adjust the scale of the graph depending on if 99/99.9 are selected. As it is the scale is distorted even if they are deselected.
> 
> By default, upon landing on the page, the scale is set up so that it matches the violin plot range (this was @rafrombrc's request.) Since by 99/99.9 are deselected by default, the scale is kept as is, but selecting them should readjust the scale accordingly.

Yes, hmm, I think the scale actually works how I thought it should-- there just seems to be some weird padding by default which made me think that the scale was incorporating values from the 99/99.9th percentiles. I see this in production too, if it hasn't been a problem yet, it's probably not worth worrying about.

> > I think we ought to add a help bubble by the "percentiles" to give a user an idea of what they mean (and also draw attention to the fact that they can be changed). This could be done as a follow-up, of course.
> 
> We do tooltip the percentiles. Do you think we should add more to it?

Oh nice! I'd probably still add some kind of basic explainer on what percentiles are as well, since this isn't super discoverable. Filed: https://github.com/mozilla/glam/issues/1629

wlach (approved):
LGTM, might be good to do a few rounds of testing on stage to see if everything works as expected.

Issue #1581, opened 2021-09-16
Consider using Glean Dictionary to search and display data relevant to Glean probes (e.g. for Firefox for Android)


wlach:
I might turn this into a proposal but filing an issue just to get the ball rolling.

Right now we use/store the [probe search service](https://github.com/mozilla/probe-search) to store and retrieve metric data.
This made some sense at the time GLAM was been developed (which was before the Glean dictionary) but the Glean Dictionary provides a much more sophisticated metadata model which includes:

* (fenix only) Tags to aid with metric discovery and use (e.g. https://dictionary.telemetry.mozilla.org/apps/fenix?page=1&search=Search)
* Annotations and commentary indicated how a metric should be interpreted (https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/browser_search_in_content)
* Links out to Looker explores relevant to the metric (especially useful if the metric is not visible in GLAM e.g. #1490, #1566)

For metrics themselves, we can probably just request the static resources in the Glean Dictionary directly. E.g. https://dictionary.telemetry.mozilla.org/data/fenix/metrics/data_addons_installed_addons.json

To return a set of results in response to a search, I have an experimental version of the Glean Dictionary which allows full text searching over the set of indexed metrics here:

https://github.com/mozilla/glean-dictionary/pull/834

Preliminary testing indicates it returns approximately the same results as the probe search service:

https://deploy-preview-834--glean-dictionary-dev.netlify.app/.netlify/functions/search?search=addons
https://dev.probe-search.nonprod.dataops.mozgcp.net/glean?limit=30&select=name%2Cdescription%2Ctype%2Cinfo&type=neq.event&or=%28name.eq.addons%2Csearch.plfts%28simple%29.addons%2Cdescription.phfts%28english%29.addons%2Cname.ilike.*addons*%29&not.and=%28type.eq.scalar%2Cinfo-%3Ecalculated-%3Elatest_history-%3Edetails-%3E%3Ekind.eq.string%29&order=info-%3Ecalculated-%3E%3Eactive.desc&product=eq.fenix

robhudson:
Would we change the flow so users start on Glean Dictionary with a link out to GLAM? Or vice versa? Or some other UI change I'm not clear on here?

wlach:
> Would we change the flow so users start on Glean Dictionary with a link out to GLAM? Or vice versa? Or some other UI change I'm not clear on here?

[...]

PR #1625, opened 2021-10-07
Use Glean Dictionary to fetch Fenix metadata

wlach:
At present, this only includes better updates on status, but there's
lots of room to extend this further.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -34,6 +34,8 @@ export default {
       __BASE_DOMAIN__: production ? '' : 'http://localhost:8000',
       __GA_TRACKING_ID__: process.env.GA_TRACKING_ID,
       __BASE_SEARCH_DOMAIN__: SEARCH_DOMAIN,
+      __GLEAN_DICTIONARY_DOMAIN__:
+        'https://deploy-preview-882--glean-dictionary-dev.netlify.app',
     }),
     string({ include: 'src/**/*.tpl' }),
     svelte({
--- a/src/routing/wrappers/Layout.svelte
+++ b/src/routing/wrappers/Layout.svelte
@@ -14,9 +14,11 @@
   $: isProbeDetailsView = $store.route.section === 'probe';
 
   const defaultTitle = 'GLAM: Glean Aggregated Metrics Explorer';
+  // Use the proper probe name if available (for Glean applications),
+  // falling back to the probeName state if not
   $: title =
     $store.route.section === 'probe'
-      ? `${$store.probeName} | GLAM`
+      ? `${$store.probe.name || $store.probeName} | GLAM`
       : defaultTitle;
 </script>
 
--- a/src/state/api.js
+++ b/src/state/api.js
@@ -56,49 +56,75 @@ export async function getProbeData(params) {
   return data;
 }
 
+function getFormattedSearchURL(
+  queryString,
+  exactSearch,
+  product,
+  resultsLimit
+) {
+  const URLResult = new URL('__BASE_SEARCH_DOMAIN__');
+
+  const params = new URLSearchParams();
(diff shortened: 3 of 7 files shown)


wlach:
This is related to #1581, though stops short of implementing everything described in that issue.

rollup.config.js
```diff
@@ -34,6 +34,8 @@ export default {
       __BASE_DOMAIN__: production ? '' : 'http://localhost:8000',
       __GA_TRACKING_ID__: process.env.GA_TRACKING_ID,
       __BASE_SEARCH_DOMAIN__: SEARCH_DOMAIN,
+      __GLEAN_DICTIONARY_DOMAIN__:
+        'https://deploy-preview-882--glean-dictionary-dev.netlify.app',
```
wlach, line 38: We should set this to https://dictionary.telemetry.mozilla.org before landing. Need mozilla/glean-dictionary#882 and a new release

wlach:
This fixes some cases where we were incorrectly reporting a metric as active. Less important but still desirable is that we now display the metrics correctly (before we replaced the category/name seperator with an underscore). For example...

Before:

image

After:

image

https://glam.telemetry.mozilla.org/fenix/probe/avif_dav1d_decode_error/explore?aggType=avg

src/state/api.js
```diff
@@ -56,49 +56,75 @@ export async function getProbeData(params) {
...
+
+  const url = useGleanDictionary
+    ? `__GLEAN_DICTIONARY_DOMAIN__/data/${product}/metrics/data_${probeName}.json`
+    : getFormattedSearchURL(probeName, true, product);
+
+  return fetch(url).then((r) => {
+    // the probe search service returns an array of results (one in this case),
+    // whereas the Glean Dictionary returns a single JSON dictionary
+    if (r.ok)
+      return r.json().then((json) => (useGleanDictionary ? json : json[0]));
```
Iinh, line 127: Why is it `json[0]` if firefox desktop?

src/routing/wrappers/Layout.svelte
```diff
@@ -14,9 +14,11 @@
   $: isProbeDetailsView = $store.route.section === 'probe';
 
   const defaultTitle = 'GLAM: Glean Aggregated Metrics Explorer';
+  // Use the proper probe name if available (for Glean applications),
+  // falling back to the probeName state if not
   $: title =
     $store.route.section === 'probe'
-      ? `${$store.probeName} | GLAM`
+      ? `${$store.probe.name || $store.probeName} | GLAM`
```
Iinh, line 21: I'm wondering if using the new metric name format (e.g. old: `avif_dav1d_decode_error` vs new: `avif.dav1d_decode_error`) will affect the `View SQL Query` feature, since it seems like we're using the old format for our queries? 

What do you think @robhudson, if you're available?

Iinh (approved):
This looks really great! Thank you for your work @wlach.

src/state/api.js
wlach: See the comment above on lines 124-125: the probe search service returns an array (with a single result)

Iinh: oops, sorry I missed that!

src/routing/wrappers/Layout.svelte
wlach: Good point, we need to rewrite this probe name. Easily done.

wlach:
Thanks @Iinh! Let's wait until Monday to merge this.

wlach:
I did up a quick new release of the Glean Dictionary with CORS enabled on `/data/`:

https://github.com/mozilla/glean-dictionary/releases/tag/v1.14.0

So we can land this now.

PR #1649, opened 2021-10-13
Show 99 and 99.9 percentiles in table view

Iinh:
Now that we added 99 and 99.9 to the list of available percentiles, let's also show them in the table view.

### DIFF
--- a/src/components/table/ProbeTableView.svelte
+++ b/src/components/table/ProbeTableView.svelte
@@ -8,6 +8,7 @@
     gatherProbeKeys,
     gatherAggregationTypes,
   } from '../../utils/probe-utils';
+  import { PERCENTILES } from '../../utils/constants';
 
   export let data;
   export let probeType = 'categorical';
@@ -70,9 +71,7 @@
     data={selectedData}
     {aggregationLevel}
     colorMap={probeType === 'categorical' ? colorMap : percentileLineColorMap}
-    visibleBuckets={probeType === 'categorical'
-      ? visibleBuckets
-      : [5, 25, 50, 75, 95]}
+    visibleBuckets={probeType === 'categorical' ? visibleBuckets : PERCENTILES}
     keyFormatter={probeType === 'categorical' ? (v) => v : (v) => `${v}%`}
     valueFormatter={probeType === 'categorical'
       ? formatPercentDecimal
--- a/src/components/table/TableView.svelte
+++ b/src/components/table/TableView.svelte
@@ -47,6 +47,12 @@
     'labeled_counter',
   ]);
 
+  const PERCENTILES = [5, 25, 50, 75, 95, 99, 99.9];
+
+  visibleBuckets = METRICS_WITH_PERCENTILE_DATA.has(data[0].metric_type)
+    ? PERCENTILES
+    : visibleBuckets;
+
   let largestAudience;
   $: largestAudience = Math.max(...data.map((d) => d.audienceSize));
 </script>
--- a/src/components/table/TableView.svelte
+++ b/src/components/table/TableView.svelte
@@ -18,7 +18,6 @@
   } from '../../utils/formatters';
 
   import { backwards } from '../../utils/iterables';
-  import { METRICS_WITH_PERCENTILE_DATA } from '../../utils/constants';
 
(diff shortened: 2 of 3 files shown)


src/components/table/TableView.svelte
```diff
@@ -47,6 +47,12 @@
     'labeled_counter',
   ]);
 
+  const PERCENTILES = [5, 25, 50, 75, 95, 99, 99.9];
+
+  visibleBuckets = METRICS_WITH_PERCENTILE_DATA.has(data[0].metric_type)
+    ? PERCENTILES
+    : visibleBuckets;
```
wlach: IMO it's not ideal to redefine input into a component.

I'm feeling like it makes more sense to do this in the higher level, containing component which already has a fair amount of business logic and keep this one simpler: https://github.com/mozilla/glam/blob/main/src/components/table/ProbeTableView.svelte#L69

There's already a definition of percentiles there. Might make sense to move `METRICS_WITH_PERCENTILE_DATA` to `src/constants.js` while you're at it.

wlach (changes requested):
Have some minor suggestions here.

src/components/table/TableView.svelte
Iinh: Thanks @wlach! This is now addressed with the latest commit.

src/components/table/ProbeTableView.svelte
```diff
@@ -70,9 +71,7 @@
     data={selectedData}
     {aggregationLevel}
     colorMap={probeType === 'categorical' ? colorMap : percentileLineColorMap}
-    visibleBuckets={probeType === 'categorical'
-      ? visibleBuckets
-      : [5, 25, 50, 75, 95]}
+    visibleBuckets={probeType === 'categorical' ? visibleBuckets : PERCENTILES}
```
wlach, line 74: This doesn't seem *quite* equivalent to your previous commit -- does non-categorical imply `METRICS_WITH_PERCENTILE_DATA`? https://github.com/mozilla/glam/pull/1649/commits/72ef1b9d8ab36241646f039599dafb545c7112d0

Iinh: good catch! let me dig deeper to see if non-categorical is actually equivalent to `METRICS_WITH_PERCENTILE_DATA` 🤔

Iinh: So I was hesitant to use `probeType 'categorical'` as a condition to check whether a metric has percentile data in #1599 because I thought probeType was from viewType, and as defined here:

https://github.com/mozilla/glam/blob/a6b2802ebb3d70ec9372efecd466eb1e298625de/src/routing/wrappers/Probe.svelte#L33

https://github.com/mozilla/glam/blob/a6b2802ebb3d70ec9372efecd466eb1e298625de/src/config/fenix.js#L113-L116

Initially I thought using `viewType` from the above code snippet can't determine whether a metric has percentile or not (my fault!) Upon further checking I realized that there are actually 2 viewType values: one in `$store`, and one in `$dataset` (which is actually the value passed to probeType):

https://github.com/mozilla/glam/blob/a6b2802ebb3d70ec9372efecd466eb1e298625de/src/config/fenix.js#L135

This is a long explanation to answer that yes, non-categorical does imply `METRICS_WITH_PERCENTILE_DATA` (well, actually non-categorical covers more metric types than `METRICS_WITH_PERCENTILE_DATA`, which is good bc it should cover all the percentile metrics that I missed in the last PR 🤦🏻‍♀️)

src/components/table/TableView.svelte
```diff
@@ -79,7 +69,7 @@
       {currentPage} />
   </div>
 
-  {#if METRICS_WITH_PERCENTILE_DATA.has(data[0].metric_type)}
+  {#if bucketTypeLabel === 'percentiles'}
```
Iinh, line 72: My bad, I didn't realize that we already defined `bucketTypeLabel` to handle the use case of differentiating whether a metric has categorical or percentile data.

https://github.com/mozilla/glam/blob/a6b2802ebb3d70ec9372efecd466eb1e298625de/src/components/table/ProbeTableView.svelte#L84-L86

wlach (approved):
🎉

PR #1650, opened 2021-10-13
Make it clear when a button in table view is selected

Iinh:
Address another feedback of @ecsmyth.

**Before**

CleanShot 2021-10-13 at 09 57 19@2x

**After**

CleanShot 2021-10-13 at 09 56 06@2x

Note: I'll also ask for feedback from @ecsmyth in the internal channel later (can't add him as a reviewer in this PR for some reason).

### DIFF
--- a/src/components/table/TableView.svelte
+++ b/src/components/table/TableView.svelte
@@ -87,7 +87,7 @@
           on:click={() => {
             showHistogramData = false;
           }}
-          level="medium"
+          level={showHistogramData ? 'medium' : 'small'}
           compact>
           Percentile Data
         </Button>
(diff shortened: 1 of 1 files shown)


wlach (approved):

Iinh:
> the image makes it look like the text color for the selected button changes from the dark blue to white. Is that correct? If so, I think that needs to change. The lack of contrast with the button background color makes it hard to read. It also seems to differ from other buttons that retain a dark font color for the selected button.

@ecsmyth Thanks for your feedback! I have changed the selected button styling to look more like our other buttons on the page.

CleanShot 2021-10-14 at 11 28 54

ecsmyth (approved):

PR #1638, opened 2021-10-12
Revert "Bump google-cloud-bigquery from 2.28.0 to 2.28.1 (#1627)"

Iinh:
Revert back to google-cloud-bigquery 2.28.0 to fix this deployment error: https://app.circleci.com/pipelines/github/mozilla/glam/4063/workflows/be5d5be6-640a-459b-9c5f-b23f24489b69/jobs/7935.

This reverts commit 520535d6710af38f6e22beb74d615257d0e500a4.

PR #1651, opened 2021-10-13
Revert "Bump python from 3.9-slim to 3.10.0-slim (#1618)"

Iinh:
Initially from reading the error log I thought that the cause for publish-latest test failing was due to google-cloud-bigquery upgrade:

CleanShot 2021-10-13 at 10 17 53@2x

but reverting that commit in #1638 didn't fix it. Through tracing the steps it seems that test started failing after upgrading Python in commit 7655d08d83c2a331b41ee694e9739191f0813393, so we should revert this commit instead.

I looked at @wlach's prior art https://github.com/mozilla/lookml-generator/pull/108/ to see if there's a way to prevent this going forward but it looks like glam Docker versioning is already tight?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # BACKEND IMAGE
-FROM python:3.10.0-slim AS backend
+FROM python:3.9-slim AS backend
 
 ENV PYTHONDONTWRITEBYTECODE=1
 ENV PYTHONUNBUFFERED=1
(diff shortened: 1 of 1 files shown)


wlach:
Oh yes, that could be the problem. Most packages haven't explicitly said they're compatible with 3.10 yet, so not surprised you're seeing errors. Just sticking with python 3.9 for now makes sense to me.

wlach (approved):

PR #1659, opened 2021-10-15
Change how we handle bug/review links

Iinh:
In the past, I believe that we passed in number as Bugzilla references for Glean metadata. This has caused some issues in GLAM, such as failure to load metadata, because GLAM currently assumes that every link passed in is a string.

For example if we look at [metrics.syncing_system](https://glam.telemetry.mozilla.org/fenix/probe/metrics_syncing_items/explore?aggType=avg), the page is blank even though we do have metadata for this metric.

CleanShot 2021-10-15 at 15 40 21@2x

Glean Dictionary handles bug/review links very well, so I copied over the 2 methods we use in Glean Dictionary to fix it.

After:

CleanShot 2021-10-15 at 15 42 29@2x

### DIFF
--- a/src/routing/pages/probe/FenixDetails.svelte
+++ b/src/routing/pages/probe/FenixDetails.svelte
@@ -4,7 +4,7 @@
   import Brackets from '../../../components/icons/Brackets.svelte';
   import { store, dataset } from '../../../state/store';
   import { downloadString } from '../../../utils/download';
-  import { extractBugId } from '../../../utils/urls';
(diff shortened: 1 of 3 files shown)


wlach (approved):
Thank you! Eventually we might want to move this logic to glean-dictionary's python script so we don't have to define this in both projects.

Iinh:
> Thank you! Eventually we might want to move this logic to glean-dictionary's python script so we don't have to define this in both projects.

Agreed! Filed: https://github.com/mozilla/glean-dictionary/issues/913

PR #1676, opened 2021-10-26
Link out to Looker from GLAM explore page

### LINKED ISSUE #1610: Add links out to Looker from GLAM metrics pages
In many cases, a Looker explore for a metric may be a useful complement to the data GLAM displays. It is also useful in the case where GLAM doesn't support a metric type (e.g. events, see #1490). For these cases, we should streamline going to Looker for a metric.

The shortest path to implementing this probably runs through #1581

Iinh:
This PR adds a link to Looker from the metric explore page using the available Looker link from the Glean Dictionary metadata. Fixes #1610.

Here is how it looks when it's enabled (when there's a Looker url):
CleanShot 2021-10-27 at 10 42 04@2x

When it's disabled (Firefox probes or when there's no url available):
CleanShot 2021-10-27 at 10 41 40@2x

(I tried to use Storybook to isolate the UI changes, but the existing Storybook needs a bit of cleanup, which will be in a separate PR.)


src/components/LookerLink.svelte
```diff
@@ -0,0 +1,79 @@
...
+
+    lookerURL =
+      variant &&
+      variant.etl.ping_data[sendInPings].looker &&
+      variant.etl.ping_data[sendInPings].looker.metric.url;
+
+    helpText =
+      (lookerURL && 'Explore this metric in Looker.') ||
+      (product === 'firefox' &&
+        'Looker is only available in Glean applications.') ||
```
wlach: I might suggest putting this inside the HTML/presentation instead of making it a reactive assignment.

```suggestion
        'Looker is only available for Glean applications.') ||
```

wlach (approved):
This looks great to me. Tagging @ecsmyth on the helptext language, as I know communication around the Glean migration is something he's thinking about a lot these days.

Iinh:
hey @ecsmyth, I'm going to merge this if no further change to the [help text](https://github.com/mozilla/glam/blob/2d42a6984baa3811e321014208287e0be2ab90ff/src/components/LookerLink.svelte?_pjax=%23js-repo-pjax-container%3Afirst-of-type%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%3Afirst-of-type%2C%20%5Bdata-pjax-container%5D%3Afirst-of-type#L23-L28) is required. I was hoping that we could get this in with the new release :)

PR #1685, opened 2021-11-02
Update package-lock.json to v2

Iinh:
[The latest tag](https://github.com/mozilla/glam/releases/tag/2021.11.0) deployment to stage failed. See: https://app.circleci.com/pipelines/github/mozilla/glam/4184/workflows/ca24fe0e-d036-4219-be26-df904b3e8961/jobs/8209

Reading the error log, it looks like the issue is that Docker wasn't able to update to the latest npm package (v8.1.2). This might be due to the fact that our current package-lock.json is outdated.

PR #1694, opened 2021-11-08
Install Python with node Docker image

Iinh:
This PR attemps to fix `publish-latest` job failing. See the error log from the last merged commit: https://app.circleci.com/pipelines/github/mozilla/glam/4184/workflows/ca24fe0e-d036-4219-be26-df904b3e8961/jobs/8209/parallel-runs/0/steps/0-105

The issue is that Docker wasn't able to build node. Initially I thought this might be due to our outdated package-lock.json file, so I updated the lockfile in #1685, but this attempt didn't work.

Upon further investigation, it looks like the reason why Docker wasn't able to build node is that it wasn't able to install `gyp`:

```
npm ERR! code 1
npm ERR! path /app/node_modules/deasync
npm ERR! command failed
npm ERR! command sh -c node ./build.js
npm ERR! gyp info it worked if it ends with ok
npm ERR! gyp info using node-gyp@8.2.0
npm ERR! gyp info using node@16.13.0 | linux | x64
npm ERR! gyp ERR! find Python 
npm ERR! gyp ERR! find Python Python is not set from command line or npm configuration
npm ERR! gyp ERR! find Python Python is not set from environment variable PYTHON
npm ERR! gyp ERR! find Python checking if "python3" can be used
npm ERR! gyp ERR! find Python - "python3" is not in PATH or produced an error
npm ERR! gyp ERR! find Python checking if "python" can be used
npm ERR! gyp ERR! find Python - "python" is not in PATH or produced an error
```

node-gyp requires Python, but our current node image -- `node:lts-slim` -- [doesn't include Python](https://github.com/nodejs/docker-node/issues/12#issuecomment-84025397). I'm suspecting that CI is failing now when it ran fine in the past might be because one of the recent dependency updates started requiring gyp. Regardless, it looks like installing Python with the node image is the way to go from now on.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,8 +56,9 @@ jobs:
       - checkout
       - run:
           command: |
-            docker-compose pull
-            docker-compose build
+            cp .env-dist .env
+            make up -v ${PWD}:/app
+            make build -v ${PWD}:/app
 workflows:
   version: 2
   test-and-deploy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,12 +48,25 @@ jobs:
             - ./node_modules
       - run: npm run build
       - run: npm run test
-
+  test-environment:
+    docker:
+      - image: cimg/base:2021.10
+    working_directory: ~/mozilla/glam/
+    steps:
+      - checkout
+      - run:
+          command: |
+            docker-compose pull
+            docker-compose build
 workflows:
   version: 2
   test-and-deploy:
     jobs:
       - frontend-tests:
+          filters:
+            tags:
+              only: /.*/
+      - test-environment:
           filters:
             tags:
               only: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,7 @@ jobs:
       - checkout
       - run:
           command: |
+            cp .env-dist .env
             docker-compose pull
             docker-compose build
 workflows:
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,8 @@ COPY glam /app/glam
 # FRONTEND BUILDER IMAGE
 FROM node:lts-slim AS frontend
 
+RUN apt-get update || : && apt-get install python -y
+
 WORKDIR /app
 
 COPY package*.json /app/
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ COPY glam /app/glam
 # FRONTEND BUILDER IMAGE
 FROM node:lts-slim AS frontend
 
-RUN apt-get update || : && apt-get install python -y
+RUN apt-get update || : && apt-get install python3 -y
 
 WORKDIR /app
 


Iinh:
cc/ @apilly

apilly (approved):
lgtm

Dockerfile
```diff
@@ -33,6 +33,8 @@ COPY glam /app/glam
 # FRONTEND BUILDER IMAGE
 FROM node:lts-slim AS frontend
 
+RUN apt-get update || : && apt-get install python -y
```
wlach:
```suggestion
RUN apt-get update || : && apt-get install python3 -y
```

`lts-slim` seems to be based on debian buster, where the python package is just actually python 2. It's python 3 we actually want.

https://hub.docker.com/_/node
https://packages.debian.org/buster/python
https://packages.debian.org/buster/python3

wlach (changes requested):
This looks fine (minus the python versioning concern), though I'm a bit concerned that we seem to have no CI testing of building this docker image, so we won't get any clear indication if something like this breaks the build.

@Iinh Could you look into adding this? You can probably model your approach after what we do in bigquery-etl. You can either do this here or in a seperate PR. The key steps are `docker-compose pull` and `docker-compose build`

https://github.com/mozilla/telemetry-airflow/blob/140d900b85c2a261f025631313fc93afe08d6055/.circleci/config.yml#L46

Iinh:
so this is still WIP, but I'm not able to run docker-compose because the container couldn't find the env file (see circleci log). I thought the checkout step would copy all files, but when I ran `ls` in the container and compared it to local dir, for some reason `checkout` didn't copy over any dot files:

```
babel.config.js     docs            package-lock.json  scripts
bin                 glam            public             setup.cfg
CHANGELOG.md        jest.config.js  pytest.ini         src
CODE_OF_CONDUCT.md  LICENSE         README.md          stories
docker-compose.yml  Makefile        requirements.in    tests
Dockerfile          manage.py       requirements.txt
dockerfiles         package.json    rollup.config.js
``` 

@wlach am I missing anything?

.circleci/config.yml
```diff
@@ -48,7 +48,16 @@ jobs:
             - ./node_modules
       - run: npm run build
       - run: npm run test
-
+  test-environment:
+    docker:
+      - image: cimg/base:2021.10
```
Iinh, line 53: I didn't use the ubuntu image ([as suggested](https://github.com/mozilla/telemetry-airflow/blob/140d900b85c2a261f025631313fc93afe08d6055/.circleci/config.yml#L47-L49)) because it gave this error:

```
Error: 
Error response from daemon: pull access denied for ubuntu-1604, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

wlach:
> so this is still WIP, but I'm not able to run docker-compose because the container couldn't find the env file (see circleci log). I thought the checkout step would copy all files, but when I ran `ls` in the container and compared it to local dir, for some reason `checkout` didn't copy over any dot files:
> ...
> @wlach am I missing anything?

Try copying `.env-dist` to `.env` per the instructions: https://github.com/mozilla/glam/blob/main/docs/development.md#configuration

It may complain about missing `DJANGO_SECRET_KEY` and `NGINX_SESSION_SECRET` keys. Since this is just for the local session, we can probably just put in some default values. This is what I did with iodide back in the day:

https://github.com/iodide-project/iodide/blob/master/.env-dist#L4

.circleci/config.yml
wlach: Yeah I think in this case using a more recent/common docker image makes sense. I don't think it really matters a ton as long as it has docker / docker-compose installed.

Iinh:
> Try copying .env-dist to .env per the instructions: https://github.com/mozilla/glam/blob/main/docs/development.md#configuration

@wlach You meant copying it as part of the CI steps for this job? I don't think it's possible, because `checkout` didn't even copy `.env-dist` or any dot files to the container.

.circleci/config.yml
```diff
@@ -48,7 +48,16 @@ jobs:
...
+    working_directory: ~/mozilla/glam/
+    steps:
+      - checkout
+      - run:
+          command: |
+            docker-compose pull
```
wlach:
```suggestion
          command: |
            cp .env-dist .env
            docker-compose pull
```

wlach:
> > Try copying .env-dist to .env per the instructions: https://github.com/mozilla/glam/blob/main/docs/development.md#configuration
> 
> @wlach You meant copying it as part of the CI steps for this job? I don't think it's possible, because `checkout` didn't even copy `.env-dist` or any dot files to the container.

You should have a full checkout when running ci. Gave a suggestion on what I think will work.

Iinh:
> It may complain about missing DJANGO_SECRET_KEY and NGINX_SESSION_SECRET keys. Since this is just for the local session, we can probably just put in some default values.

Thanks @wlach for your help. `docker-compose` didn't complain about missing keys so I figured we don't have to update `.env-dist`.

.circleci/config.yml
```diff
@@ -48,7 +48,17 @@ jobs:
...
+    docker:
+      - image: cimg/base:2021.10
+    working_directory: ~/mozilla/glam/
+    steps:
+      - checkout
+      - run:
+          command: | # bind {PWD} because we use it in docker-compose.yml
+            cp .env-dist .env
+            make up -v ${PWD}:/app
+            make build -v ${PWD}:/app
```
wlach: I would skip the `make up` for now (unless/until we have a test that validates that the full environment comes up as expected)

```suggestion
           make build -v ${PWD}:/app
```

wlach (approved):
Looks good! Using the makefile is indeed a better way of doing it than what I suggested. I just have one small comment, then I think this can land. Thanks for sticking with it!

PR #1699, opened 2021-11-11
Use non-slim nodeJS v12 for Docker front-end builder

Iinh:
Per @robhudson's comment, let's switch to using a non-slim node image. We should go with node v12 because that's the minimum version required for GLAM development: https://github.com/mozilla/glam/blob/main/docs/development.md#development

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ COPY glam /app/glam
 
 
 # FRONTEND BUILDER IMAGE
-FROM node:lts-slim AS frontend
+FROM node:12 AS frontend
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -31,7 +31,7 @@ COPY glam /app/glam
 
 
 # FRONTEND BUILDER IMAGE
-FROM node:lts-slim AS frontend
+FROM node:12 AS frontend
```
robhudson, line 34: I would just be careful here since nodejs 12 has an end of life in Apr 2022:
https://nodejs.org/en/about/releases/

But otherwise r+

robhudson (approved):

PR #1703, opened 2021-11-16
Install make with Docker node image

Iinh:


apilly (approved):

Dockerfile
```diff
@@ -34,6 +34,7 @@ COPY glam /app/glam
 FROM node:lts-slim AS frontend
 
 RUN apt-get update || : && apt-get install python3 -y
+RUN apt-get -y install make
```
robhudson, line 37: This can be combined with the above so only one layer is produced. See https://docs.docker.com/storage/storagedriver/#images-and-layers for an explanation. Not a big issue but it's nice to keep the docker layers to a minimum, if possible, especially with `RUN`, `ADD`, and `COPY` commands.

PR #1724, opened 2021-12-01
Minor improvements to probe search UX

wlach:
* Use lodash (which we were using already) instead of debounce-throttle
* Use debounce instead of throttle (this prevents us from fetching
  repeatedly while the user is still typing)
* Take advantage of that and fetch results faster (this improves
  the user experience a bit)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,8 +15,7 @@
         "d3-scale-chromatic": "3.0.0",
         "d3-shape": "3.0.1",
         "d3-time-format": "3.0.0",
-        "page": "1.11.6",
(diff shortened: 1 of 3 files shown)


wlach:
The difference is somewhat subtle, but noticeable.

Before:

https://user-images.githubusercontent.com/20569/144278612-91d9bd12-a006-41bd-aae5-492b54f031c6.mp4

After:

https://user-images.githubusercontent.com/20569/144278588-2ac40aa5-97b9-4f3c-a9d4-8e4b84598b9e.mp4

Iinh (approved):
👍

PR #1735, opened 2021-12-07
Fix memory distribution display for Fenix

wlach:
We were assuming the format of the old probe info service for fenix,
which we're no longer using. Also fix another instance of us assuming
the old format (which is harmless, because we're not using it, but still).


wlach:
Example: https://glam.telemetry.mozilla.org/fenix/probe/glean_database_size/explore?aggType=avg&app_id=release

Iinh (approved):
Ah nice catch, thanks @wlach!

PR #1754, opened 2021-12-14
 change refresh mat view to create mat view

alekhyamoz:
Seeing issues with refresh materialized view. Changing the code to create materialized view instead of refresh


glam/api/management/commands/import_desktop_aggs.py
```diff
@@ -58,6 +58,7 @@ def handle(self, bucket, *args, **options):
 
         for blob in blobs:
             # Create temp table for data.
+            print(blob)
```
emtwo: Do we need to print this for debugging? Not sure what it looks like but it might be a big thing to print?

emtwo (approved):

PR #1771, opened 2022-01-04
adding glean firefox

alekhyamoz:
Django changes required to add firefox on glean to GLAM


glam/api/management/commands/import_glean_aggs.py
```diff
@@ -47,60 +52,61 @@ def add_arguments(self, parser):
...
-        LastUpdated.objects.update_or_create(
-            product=f"fenix-{app_id}", defaults={"last_updated": timezone.now()}
-        )
-
-    def import_file(self, tmp_table, fp, app_id):
-
-        model = apps.get_model(APP_TO_MODEL[app_id])
+        for product in PRODUCT_TO_MODEL.keys():
+            self.gcs_client = storage.Client()
+            # model = apps.get_model(APP_TO_MODEL[app_id])
```
Iinh: did you mean to keep this comment?

glam/api/models.py
```diff
@@ -75,11 +75,38 @@ class Meta(AbstractFenixAggregation.Meta):
         ]
 
 
-class FenixAggregationView(AbstractFenixAggregation):
+class FenixAggregationView(AbstractGleanAggregation):
     class Meta:
         managed = False
         db_table = "view_glam_fenix_aggregation"
 
+class FirefoxAggregation(AbstractGleanAggregation):
```
Iinh: I think FOGAggregation instead of FirefoxAggregation makes it more clear that this is a Firefox on Glean product, instead of Firefox Desktop.

```suggestion
class FOGAggregation(AbstractGleanAggregation):
```

glam/api/models.py
```diff
@@ -75,11 +75,38 @@ class Meta(AbstractFenixAggregation.Meta):
...
+    class Meta(AbstractGleanAggregation.Meta):
+        db_table = "glam_firefox_aggregation"
```
Iinh:
```suggestion
        db_table = "glam_fog_aggregation"
```

glam/api/models.py
```diff
@@ -75,11 +75,38 @@ class Meta(AbstractFenixAggregation.Meta):
...
+                    "metric",
+                    "metric_type",
+                    "metric_key",
+                    "client_agg_type",
+                ],
+            )
+        ]
+
+
+class FirefoxAggregationView(AbstractGleanAggregation):
```
Iinh:
```suggestion
class FOGAggregationView(AbstractGleanAggregation):
```

glam/api/management/commands/import_glean_aggs.py
```diff
@@ -114,12 +120,12 @@ def import_file(self, tmp_table, fp, app_id):
                 """
                 cursor.copy_expert(sql, tmp_file)
 
-        log(app_id, "  Inserting data from temp table into aggregation tables.")
+        log(app_id, " Inserting data from temp table into aggregation tables.")
```
Iinh, line 121: I think you left an extra space here.
```suggestion
        log(app_id, "Inserting data from temp table into aggregation tables.")
```

glam/api/management/commands/import_glean_aggs.py
```diff
@@ -47,60 +52,61 @@ def add_arguments(self, parser):
...
+            for blob in blobs:
+                # Create temp table for data.
+                tmp_table = "tmp_import_glean_{}".format(app_id)
+                log(app_id, f"Creating temp table for import: {tmp_table}.")
+                with connection.cursor() as cursor:
+                    cursor.execute(f"DROP TABLE IF EXISTS {tmp_table}")
+                    cursor.execute(
+                        f"CREATE TABLE {tmp_table} (LIKE {model._meta.db_table})"
+                    )
+                    # The incoming CSV files don't have an `id` or `app_id`.
```
Iinh, line 71: is this comment still valid?

glam/api/management/commands/import_glean_aggs.py
```diff
@@ -21,6 +21,11 @@
     "release": "api.FenixAggregation",
 }
 
+PRODUCT_TO_MODEL = {
+    "org_mozilla_fenix" :  "api.FenixAggregation",
+    "firefox_desktop" :  "api.FirefoxAggregation",
```
Iinh:
```suggestion
    "firefox_desktop" :  "api.FOGAggregation",
```

Iinh:
Let's also update the deployment docs to mention that Glean data import now includes both Fenix and FOG data: https://github.com/mozilla/glam/blob/main/docs/development.md

glam/api/management/commands/import_glean_aggs.py
```diff
@@ -47,79 +52,79 @@ def add_arguments(self, parser):
...
-
-        LastUpdated.objects.update_or_create(
-            product=f"fenix-{app_id}", defaults={"last_updated": timezone.now()}
-        )
-
-    def import_file(self, tmp_table, fp, app_id):
-
-        model = apps.get_model(APP_TO_MODEL[app_id])
+        for product in PRODUCT_TO_MODEL.keys():
+            self.gcs_client = storage.Client()
```
robhudson: This probably doesn't need to happen in the loop... one GCS client is probably good.

glam/api/management/commands/import_glean_aggs.py
```diff
@@ -47,79 +52,79 @@ def add_arguments(self, parser):
...
+                    )
+
+                # Download CSV file to local filesystem.
+                fp = tempfile.NamedTemporaryFile()
+                log(app_id, f"Copying GCS file {blob.name} to local file {fp.name}.")
+                blob.download_to_filename(fp.name)
+
+                #  Load CSV into temp table & insert data from temp table into
+                #  aggregation tables, using upserts.
+                self.import_file(tmp_table, fp, app_id,product)
```
robhudson: Add a space after the comma, or run the `black` formatter on this file.

glam/api/management/commands/import_glean_aggs.py
```diff
@@ -47,79 +52,79 @@ def add_arguments(self, parser):
...
+        log(app_id, "Importing file into temp table.")
         with connection.cursor() as cursor:
             with open(fp.name, "r") as tmp_file:
                 sql = f"""
                     COPY {tmp_table} ({", ".join(csv_columns)}) FROM STDIN WITH CSV
                 """
                 cursor.copy_expert(sql, tmp_file)
 
-        log(app_id, "  Inserting data from temp table into aggregation tables.")
+        log(app_id, "Inserting data from temp table into aggregation tables.")
```
robhudson: I had the spaces there to logically separate the output by CSV files but feel free to remove it if you'd like. Just pointing it out just in case.

glam/api/migrations/0020_transpose_fog.py
```diff
@@ -0,0 +1,68 @@
...
+                ('histogram', models.TextField(blank=True, null=True)),
+                ('percentiles', models.TextField(blank=True, null=True)),
+            ],
+            options={
+                'db_table': 'view_glam_fog_aggregation',
+                'managed': False,
+            },
+        ),
+        
+    ]
```
robhudson: I'm not sure if you edited these by hand or not? But this migration shouldn't exist. We used to have 2 rows, one with the `histogram` and one with the `percentiles` and then we combined them. This migration was for the existing tables at the time. Since the new models already have this change, these migrations aren't needed and you should run `makemigrations` from the models.  See https://docs.djangoproject.com/en/4.0/topics/migrations/

alekhyamoz: ok. Makes sense. I will re-generate the migrations.

glam/api/management/commands/import_glean_aggs.py
alekhyamoz: Got it!

glam/api/management/commands/import_glean_aggs.py
alekhyamoz: makes sense.

glam/api/management/commands/import_glean_aggs.py
alekhyamoz: Made all the required changes.

glam/api/management/commands/import_glean_aggs.py
alekhyamoz: yes since the incoming files don't have the app_id

glam/api/management/commands/import_glean_aggs.py
alekhyamoz: removed

glam/api/models.py
alekhyamoz: made the necessary changes

robhudson (approved):
Looks great!

glam/api/management/commands/import_glean_aggs.py
Iinh: I think you forgot to add a space after the comma here. Installing `black` to lint might be a good idea, but this could be done in a follow-up, as it seems that we need to lint more than just the files in the PR.

glam/api/management/commands/import_glean_aggs.py
Iinh: oh I didn't know about that. In that case I think we should leave the spaces as they were before.

Iinh (approved):
r+ with the suggested changes addressed. This is great, thanks for working on it @alekhyamoz!

Issue #1737, opened 2021-12-08
Import FOG probe data into GLAM


alekhyamoz:
As the next step to get FOG into GLAM UI, we need to make the necessary Django script changes to import FOG CSV data (residing in GCS bucket) into Postgres table

alekhyamoz:
Created a [PR](https://github.com/mozilla/glam/pull/1771) for django changes for FOG

alekhyamoz:
The implementation has been deployed to the dev environment. 
Created a couple of tickets requesting DSREs support for testing in dev
https://mozilla-hub.atlassian.net/browse/DSRE-429
https://mozilla-hub.atlassian.net/browse/DSRE-430

Iinh:
Some notes on the front-end work required for FOG:

- First, we need to consolidate legacy telemetry and FOG search API so we only need to hit 1 API for Firefox dropdown search (this work needs to be done in the Glean Dictionary)
 - Then, use this new API for desktop search
- UI-wise, add visual treatment to the search results so users can differentiate between FOG and legacy probes

alekhyamoz:
The front-end work is being tracked in #1809

PR #1801, opened 2022-01-21
added FOG to import_glean_counts and created a model FOGcounts

alekhyamoz:
Related to issue #1737 

1. Added the FOGcounts model to the model.py and generated a new migration 0021_fog_counts.py 

2. Modified the import_glean_counts to process firefox_desktop count files.

Screenshot of sample output: 
Screen Shot 2022-01-21 at 9 54 26 AM


glam/api/views.py
```diff
@@ -277,6 +285,28 @@ def _get_fenix_counts(app_id, versions, ping_type, os, by_build):
 
     return data
 
+def _get_fog_counts(app_id, versions, ping_type, os, by_build):
```
Iinh, line 291: any reason why you decided to use this separate method instead of changing `_get_fenix_counts()` to `_get_clean_counts()`?

glam/api/views.py
```diff
@@ -219,12 +221,18 @@ def get_glean_aggregations(request, **kwargs):
...
-    elif aggregation_level == "build_id":
+    elif aggregation_level == "build_id" and product == 'fenix':
         dimensions.append(~Q(build_id="*"))
         counts = _get_fenix_counts(app_id, versions, ping_type, os, by_build=True)
+    elif aggregation_level == "build_id" and product == 'firefox':
+        dimensions.append(~Q(build_id="*"))
+        counts = _get_fog_counts(app_id, versions, ping_type, os, by_build=True)
+    elif aggregation_level == "build_id" and product == 'firefox':
+        dimensions.append(~Q(build_id="*"))
+        counts = _get_fog_counts(app_id, versions, ping_type, os, by_build=True)
```
Iinh: I think we can clean up the logic here by nesting either the `if aggregation_level ==` or `if product == ` under the other condition.

glam/api/models.py
```diff
@@ -229,6 +229,26 @@ class Meta:
                 fields=["app_id", "channel", "version", "ping_type", "build_id", "os"],
             )
         ]
+class FOGCounts(models.Model):
```
Iinh, line 232: is it possible to consolidate `FenixCounts` and `FogCounts` to just one class `GleanCounts`?

glam/api/views.py
alekhyamoz: To maintain the modularity, in case the definition of the upcoming glean apps is different, we will be able to accomodate. Just as we have separate models for Fenix and Fog

glam/api/models.py
alekhyamoz: This is again to maintain modularity. if anything changes in either of the models we will be able to accommodate them and pass along information into different tables.

glam/api/views.py
Iinh: these 2 blocks look identical?

alekhyamoz: Good catch! I will correct that

Iinh: something like this:
```
if aggregation_level == "version":
    if product == 'fenix':
        dimensions.append(Q(build_id="*"))
        counts = _get_fenix_counts(app_id, versions, ping_type, os, by_build=False)
    if product == "firefox":
        ...
elif aggregation_level == "build_id":
    if product == 'fenix':
        ...
```

although if we have one single method `_get_glean_counts` then it'd be even cleaner and we don't have to nest conditions like this.

glam/api/models.py
Iinh: that makes sense. thanks for the explanation!

glam/api/views.py
alekhyamoz: ok will make the change to the nested if-else

Iinh (approved):

PR #1808, opened 2022-01-26
Add "Select All" button for categorical histograms (fixes #1085)

### LINKED ISSUE #1085: Show all results on 'table' view of categorical histograms
We've had a user (@pbz) ask in Slack to be able to see the long tail of buckets in a [categorical histogram](https://glam.telemetry.mozilla.org/firefox/probe/external_protocol_handler_dialog_context_scheme/explore). (We're currently only showing the top 10 and then combining the rest into "other".)

Fitting all of the categories onto the graph would be tough, but it'd be easy to display them all on the 'table' view, at least.

Iinh:
This PR makes 2 UI changes:

1) A "Select All" button to the category dropdown menu

- enabled:
CleanShot 2022-02-09 at 03 29 23@2x

- disabled:
CleanShot 2022-02-09 at 03 29 41@2x

2) Include the dropdown selector on the Table view page
CleanShot 2022-02-09 at 10 44 57@2x

Some example probes to test out locally:
- https://glam.telemetry.mozilla.org/firefox/probe/avif_cicp_mc/explore?activeBuckets=%5B%22BT601%22%2C%22BT709%22%2C%22IDENTITY%22%2C%22UNSPECIFIED%22%2C%22RESERVED%22%2C%22FCC%22%2C%22BT470BG%22%2C%22SMPTE240%22%2C%22YCGCO%22%2C%22BT2020_NCL%22%5D
- https://glam.telemetry.mozilla.org/firefox/probe/external_protocol_handler_dialog_context_scheme/explore?activeBuckets=%5B%22magnet%22%2C%22OTHER%22%2C%22MICROSOFT_APP%22%2C%22mailto%22%2C%22TELEGRAM%22%2C%22tel%22%2C%22steam%22%2C%22whatsapp%22%2C%22ftp%22%2C%22vscode%22%5D&process=parent

Fixes #1085.

### DIFF
--- a/src/components/explore/CategoricalMenu.svelte
+++ b/src/components/explore/CategoricalMenu.svelte
@@ -1,0 +1,152 @@
+<script>
+  import { OptionMenu, Option, OptionDivider } from '@graph-paper/optionmenu';
+
+  import { store } from '../../state/store';
+
+  import KeySelectionControl from '../controls/KeySelectionControl.svelte';
+  import ColorSwatch from '../controls/ColorSwatch.svelte';
+
+  import { Button } from '@graph-paper/button';
+
+  import { numericStringsSort } from '../../utils/sort';
+  import { numHighlightedBuckets } from '../../config/shared';
+
+  export let data;
+  export let activeBuckets;
+  export let bucketColorMap;
+  export let bucketOptions;
+
+  export let bucketSortOrder = (a, b) => (a < b ? 1 : -1);
+
+  function makeSelection() {
+    return function onSelection(event) {
+      store.setField('activeBuckets', event.detail.selection);
+    };
+  }
+
+  let showOptionMenu = false;
+  let coloredBuckets = [];
+  let everActiveBuckets = [];
+  let sortedImportantBuckets = [];
+  let sortedUnimportantBuckets = [];
+
+  if (bucketOptions.length > numHighlightedBuckets) {
+    showOptionMenu = true;
+    const lastDataset = data[data.length - 1];
+
+    coloredBuckets = Object.entries(lastDataset.counts)
+      .sort(([bucketAName, bucketAValue], [bucketBName, bucketBValue]) => {
+        const bucketValueDifference = bucketAValue - bucketBValue;
+        if (bucketValueDifference === 0) {
+          return bucketBName - bucketAName;
+        }
+        return bucketValueDifference;
+      })
+      .slice(-numHighlightedBuckets)
+      .map(([bucket]) => bucket);
+  }
+
+  let selectAllCategories = false;
+
+  $: if (showOptionMenu) {
+    everActiveBuckets = [...new Set([...activeBuckets, ...everActiveBuckets])];
+
+    // An important bucket is any bucket that:
+    //
+    //   (a) is colored
+    //   (b) is currently active
+    //   (c) has been active at some point in the past
+    //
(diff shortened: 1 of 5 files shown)


src/components/explore/CategoricalMenu.svelte
```diff
@@ -0,0 +1,152 @@
...
+  <label class="body-control-set--label">Categories</label>
+  {#if showOptionMenu}
+    <OptionMenu
+      multi
+      on:selection={(evt) => {
+        store.setField('activeBuckets', evt.detail.keys);
+      }}>
+      <div style="margin-left: 4.5em">
+        <Button
+          tooltip={'select all available categories'}
```
brizental: The tooltip is redundant with the text of the button.

brizental (changes requested):
Codewise I think this looks good. I have a few issues with the interface:

1. The contrast on the disabled button is not very good, I can't read the text.
2. The disabled button should be really disabled. Right now I can click it and it will close the dropdown, that is unexpected.
3. This is a nitpick, but I find the positioning and style of the button a bit weird for it's placement I feel it would make more sense as a full wdith button, like so: 
Screenshot 2022-02-17 at 12 51 00
Feel free to ignore this nit as you see fit :)

Iinh:
Thanks so much for the review Bea!

> The contrast on the disabled button is not very good, I can't read the text.

> This is a nitpick, but I find the positioning and style of the button a bit weird for it's placement I feel it would make more sense as a full wdith button, like so:

I changed the button design to be full width like you suggested, and I agree it looks better! Also, this new design fixes the contrast issue with the disabled button.

CleanShot 2022-02-23 at 09 42 04@2x

CleanShot 2022-02-23 at 09 42 13@2x

> The disabled button should be really disabled. Right now I can click it and it will close the dropdown, that is unexpected.

This is probably because we use a graph-paper Button component so it was not really disabled. Switching to using a regular button element seems to address this issue.

src/components/explore/CategoricalMenu.svelte
```diff
@@ -0,0 +1,160 @@
...
+      .filter((bucket) => !sortedImportantBuckets.includes(bucket))
+      .sort(numericStringsSort);
+  }
+
+  function handleSelectAllCategories() {
+    store.setField('activeBuckets', bucketOptions);
+  }
+
+  $: selectAllCategories =
+    $store.activeBuckets.length === bucketOptions.length && true;
```
brizental: Why is `&& true` necessary here?

brizental (approved):
I found two more nits with the menu upon a second look. 

1. When the screen in smaller the menu height gets way too small:
Screenshot 2022-02-25 at 15 18 48
2. I feel i'ts highly annoying that the menu closes everytime an option is selected / unselected, is there anything we can do about it? Maybe something to fix upstream?

None of this is blocking and can be done as follow-ups, so r+ :)

src/components/explore/CategoricalMenu.svelte
Iinh: oops this is not right! thanks for catching this!

Iinh:
> I feel i'ts highly annoying that the menu closes everytime an option is selected / unselected, is there anything we can do about it? Maybe something to fix upstream?

Indeed this requires an upstream change, specifically the `graph-paper` <Option> component. I'll open an issue about this later, thanks for the input @brizental!

PR #1819, opened 2022-02-03
Use Glean Dictionary updated search API

### LINKED ISSUE #1761: Should use Glean Dictionary's "official" search API
In #1736 we swapped in the Glean Dictionary as the source of probe/metric metadata. However, it's using the weird `.netlify` API endpoint. Instead we should use the new API search endpoints landed in https://github.com/mozilla/glean-dictionary/pull/1028.

Should just require a change small here to update the path:

https://github.com/mozilla/glam/blob/a075a955ab8626565bc1fa28e919a92a351edecb/src/state/api.js#L61

Iinh:
Fixes #1761.

With the new Glean Dictionary release, we can now start using the new official search API. 

E.g: https://dictionary.telemetry.mozilla.org/api/v1/metrics_search_firefox_legacy?search=video_dropped

### DIFF
--- a/src/state/api.js
+++ b/src/state/api.js
@@ -58,7 +58,7 @@ export async function getProbeData(params) {
 
 function getProbeSearchURL(productId, queryString, resultsLimit) {
   const URLResult = new URL(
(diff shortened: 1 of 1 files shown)


alekhyamoz (approved):
Looks good

PR #1846, opened 2022-02-22
Fix failed front-end tests

Iinh:
We have some linting issues that somehow escaped the front-end check in the last commit.


edugfilho (approved):
lgtm. thank you!

PR #1853, opened 2022-02-28
create sample counts migration and management commands for both firef…

### LINKED ISSUE #1240: Add sample counts
On TMO's evolution view the bottom chart shows the sample counts. The same could be done for GLAM. 

An example of how sample counts are meaningful was reported in the glam channel on Slack:
> I'm looking at telemetry for minor GCs and this shows us how many were performed.  Changing the GC scheduling affects how many GCs are run as well as how long they individually take, and the overall impact depends on both.

alekhyamoz:
…ox desktop and glean products

This is related to issue #1240. 
The ETL is now processing the sample counts data, available at 
https://console.cloud.google.com/storage/browser/moz-fx-data-glam-prod-fca7-etl-data;tab=objects?cloudshell=false&project=moz-fx-data-glam-prod-fca7&supportedpurview=project&prefix=&forceOnObjectsSortingFiltering=false&pli=1&rapt=AEjHL4M_OFRxeze4m0D95kqm0-bfnoqkRnv5wHKOrE4pjS75KF3QXRTjTIQlcsBH70ASQzHuv0wq0aAI9zSSRw9R0G0Vi4iJKQ

This PR includes the necessary migration and management commands to pull the data into Postgres tables. 

Models and migrations have been written to create separate sample_counts table will be created in postgres for each application i.e. firefox_desktop_sample_counts, fog_sample_counts, fenix_sample_counts. 

The management commands import_desktop_sample_counts.py and import_glean_sample_counts.py are the runtime commands (run via Airflow) that will inject the GCS data into Postgres. 
`
Commands to run`
```
./manage.py makemigrations
./manage.py migrate
./manage.py import_desktop_sample_counts
./manage.py import_glean_sample_counts

```
We should see the new tables (glam_firefox_sample_counts, glam_fenix_sample_counts, glam_fog_sample_counts) created and the data from prod GCS buckets imported to the sample counts table


Iinh:
Looks like we still need to rebase this branch. Try `git fetch`, and then `git pull --rebase origin main` before force pushing it.

glam/api/management/commands/import_desktop_sample_counts.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--bucket",
+            default=constants.GCS_BUCKET,
+            help="The bucket location for the exported counts",
+        )
+
+    def handle(self, bucket, *args, **options):
+        print("Im here  ....")
```
edugfilho: Looks like a leftover from debugging?
Given that there's the `log` function, maybe the other prints in this function as well?

alekhyamoz: oh yes, will remove the print statement

edugfilho (dismissed):

glam/api/migrations/0022_add_sample_counts.py
```diff
@@ -0,0 +1,105 @@
...
+            name='FirefoxSampleCounts',
+            fields=[
+                ('id', models.AutoField(primary_key=True, serialize=False)),
+                ('channel', models.IntegerField(blank=True, choices=[(1, 'nightly'), (2, 'beta'), (3, 'release')], null=True)),
+                ('version', models.CharField(blank=True, max_length=100, null=True)),
+                ('build_id', models.CharField(max_length=100)),
+                ('os', models.CharField(max_length=100)),
+                ('process', models.CharField(max_length=100)),
+                ('metric', models.CharField(max_length=100)),
+                ('key', models.CharField(max_length=100)),
```
Iinh, line 66: The order of these 2 columns does not seem right. According to [this PR](https://github.com/mozilla/bigquery-etl/pull/2751/files), `key` should be before `metric`.

glam/api/migrations/0022_add_sample_counts.py
```diff
@@ -0,0 +1,105 @@
...
+                ('client_agg_type', models.CharField(max_length=100)),
+                ('total_sample', models.BigIntegerField(blank=True, null=True)),
```
Iinh, line 68: tiny nit: I think calling this column `sample_count` is better than `total_sample` (however I understand if you chose this name because you want to keep it consistent with `total_users` for user count)

glam/api/migrations/0022_add_sample_counts.py
```diff
@@ -0,0 +1,105 @@
...
+            ],
+            options={
+                'db_table': 'glam_fenix_sample_counts',
+            },
+        ),
+        migrations.CreateModel(
+            name='FirefoxSampleCounts',
```
Iinh, line 57: nit: I think it would be clearer if this was named `DesktopSampleCounts`, rather than generic Firefox (as they are all Firefox products)

glam/api/management/commands/import_desktop_sample_counts.py
```diff
@@ -0,0 +1,98 @@
...
+FILENAME = os.path.basename(__file__).split(".")[0]
+
+
+def log(message):
+    print(f"{datetime.datetime.now().strftime('%x %X')} - {FILENAME} - {message}")
+
+
+class Command(BaseCommand):
+
+    help = "Imports sample counts"
```
Iinh, line 23: This file, and `import_glean_sample_counts`, are almost identical to `import_user_counts.py` and `import_glean_counts.py`. Would it make more sense to consolidate adding sample and user counts into just one command, e.g `import_desktop_counts` for Desktop and `import_glean_counts` for Glean?

glam/api/models.py
```diff
@@ -226,7 +226,7 @@ class Meta:
         constraints = [
             models.UniqueConstraint(
                 name="fenix_counts_unique_dimensions",
-                fields=["app_id", "channel", "version", "ping_type", "build_id", "os"],
+                fields=["app_id", "channel", "version", "ping_type", "build_id", "os",],
```
Iinh, line 229: we should take out the extra comma after "os" here.

glam/api/models.py
```diff
@@ -248,6 +248,72 @@ class Meta:
...
+
+    class Meta:
+        db_table = "glam_fenix_sample_counts"
+        constraints = [
+            models.UniqueConstraint(
+                name="fenix_sample_counts_unique_dimensions",
+                fields=["app_id", "channel", "version", "ping_type", "build_id", "os","key","metric","client_agg_type"],
+            )
+        ]
+class FOGSampleCounts(models.Model):
```
Iinh, line 296: we probably want to have some space between classes here, like what you did in lines 273-274. I wonder if we should enforce a linter on these files so we don't have to manually check for formatting inconsistencies like this.

Iinh (changes requested):
After running:
```
./manage.py migrate
./manage.py import_desktop_sample_counts
./manage.py import_glean_sample_counts
```
I was able to see the new tables created and populated with data from GCS. However, after reviewing the tables, I have 2 upstream suggestions:

1) Add `sample_count` column to the aggregate tables.

ClientCountOverTime and ClientComparison graph use data from `total_users` column from `glam_*_aggregation` tables. To make adding sample count to frontend easier, we should have this data in the same table (just like what we're doing with user count), or it would require substantial code change frontend-wise. Is this something we can do on the ETL side?

2) Simplify sample count table schema

If we can make (1) happen, I believe the data needed for the sample count tables will be greatly reduced. The information currently stored for sample count tables, especially on the desktop side, is too big. Also, most of the columns seem to be redundant as we already have them in the aggregate tables. 

For more context, we use the data in the user count table `glam_firefox_counts` to [calculate the total distinct client IDs for  `total_addressable_market`](https://github.com/mozilla/glam/blob/83fe3dfe3ecca3ab51258434125b044ac5365f25/glam/api/views.py#L136). The goal for that table is for us to do a quick query when the aggregation_level is changed, so it should be as small as possible. As we're going to do the same for sample count, I suggest that we keep schema for sample count similar to the [current user count tables](https://github.com/mozilla/glam/blob/83fe3dfe3ecca3ab51258434125b044ac5365f25/glam/api/models.py#L193-L201) (unless I missed anything important etl-wise).

Thanks @alekhyamoz for working on this!

edugfilho:
I'm confused with the result of the following query: It looks like the column `key` actually contains `metrics` and vice-versa?
Also, there are lots of blank metrics (hence the `WHERE` clause for clearer results) which are, I guess, expected for keys.
Screen Shot 2022-03-01 at 5 56 08 PM

glam/api/management/commands/import_desktop_sample_counts.py
alekhyamoz: I did think about this. My rationale behind separating them was we could run them independently (eliminating the dependency on each other). We could turn on/off each one separately

glam/api/migrations/0022_add_sample_counts.py
edugfilho: This probably explains [my comment](https://github.com/mozilla/glam/pull/1853#issuecomment-1055943886)

PR #1855, opened 2022-03-02
Add FOG metrics to frontend

### LINKED ISSUE #1809: Front-end (UI) work required for FOG
The next step after importing FOG data is to make the data visible via GLAM UI

Iinh:
Fixes #1809.

Using the search feature will trigger just one API call to the Glean Dictionary for Fx desktop (depends on https://github.com/mozilla/glean-dictionary/pull/1163.)

There is a minor URL change:

- ~~`/legacy` path for legacy telemetry (used to be `/firefox`)~~ Decided to stick to `/firefox` to avoid confusion from users
- `/fog` for FOG metrics

UI wise, to differentiate between legacy and FOG probes, FOG results will look like this:

CleanShot 2022-04-04 at 21 03 26@2x

The FOG Explore page, for the most part will look like the Fenix page, with some minor differences (OS choices are similar to desktop, and there's only one ping type "All")

CleanShot 2022-04-04 at 21 08 00@2x


Iinh:
Using the search feature will trigger just one API call to the Glean Dictionary for Fx desktop. There is a minor URL change:

- `/legacy` path for legacy telemetry (used to be `/desktop`)
- `/fog` for FOG metrics

UI wise, to differentiate between legacy and FOG probes, FOG results will look like this:

CleanShot 2022-04-04 at 21 03 26@2x

The FOG Explore page, for the most part will look like the Fenix page, with some minor differences (OS choices are similar to desktop, and there's only one ping type "All")

CleanShot 2022-04-04 at 21 08 00@2x

Iinh:
Change 'Firefox on Glean' to just 'Glean' in the search results: 

CleanShot 2022-04-12 at 11 25 17@2x

rollup.config.js
```diff
@@ -23,7 +23,8 @@ export default {
     replace({
       __BASE_DOMAIN__: production ? '' : 'http://localhost:8000',
       __GA_TRACKING_ID__: process.env.GA_TRACKING_ID,
-      __GLEAN_DICTIONARY_DOMAIN__: 'https://dictionary.telemetry.mozilla.org',
+      __GLEAN_DICTIONARY_DOMAIN__:
```
edugfilho: My understanding is this uri calls the api for unified probes. But it's a temporary one, right? 
If so, do we have an issue or another reminder to track the update to the permanent one when it's available?

rollup.config.js
```diff
@@ -23,7 +23,8 @@ export default {
...
+        'https://deploy-preview-1163--glean-dictionary-dev.netlify.app',
```
Iinh: This needs to be changed back to `https://dictionary.telemetry.mozilla.org` before merging.

src/components/search/SearchResults.svelte
```diff
@@ -91,6 +91,31 @@
...
+      store.setField('product', 'fenix');
+      store.setField('productDimensions', {
+        app_id: 'nightly',
+        os: 'Android',
+        ping_type: 'metrics',
+        aggregationLevel: 'build_id',
+      });
+      return 'fenix';
+    }
+    return 'firefox';
```
edugfilho: we're defaulting to `firefox` here. Is it something that can bite us later? If so maybe we should add another clause to be sure we're dealing with `firefox`. 
In case we go this way and the code starts to look too complicated, we can also consider a `switch-case` statement, with `default` throwing an `Error`

src/state/store.js
```diff
@@ -24,22 +24,28 @@ function getDefaultState(
...
+    state.product = 'firefox';
   }
 
   state.probeName = '';
   state.ref = getFromQueryString('ref') || '';
   state.hov = getFromQueryString('hov') || '';
   state.timeHorizon = getFromQueryString('timeHorizon') || 'MONTH';
   state.route = {};
-  state.searchProduct = state.product || 'firefox';
+  state.searchProduct = searchProducts[state.product] || 'firefox';
```
edugfilho: Is the fallback to `firefox` here redundant since we have `state.product` defaulting to `firefox` on L36 already?

edugfilho (commented):
Looks great to me in general. I wish I was more acquainted with this code to be able to provide deeper feedback.
My questions are all minor, so if you want to resolve them without changing code, I'll be 100% ok and will approve

rollup.config.js
edugfilho: > This needs to be changed back to `https://dictionary.telemetry.mozilla.org` before merging.

I just saw this /\

Slightly change in my question, then: 
- Does the Glean Dictionary prod endpoint already support `fog_and_legacy`? If not, I think we just have to make sure to deploy this feature after it does, right? If so, do we have something to help us remember that?

Iinh: I just did a production release for the Glean Dictionary which includes the consolidated search API change. We can update it here now.

src/state/store.js
Iinh: you're right!

src/components/search/SearchResults.svelte
Iinh: I added another clause to capture the `firefox` case. This function now returns `undefined` if no condition is met, which I think is more semantic for Javascript :)

edugfilho (approved):
Great stuff!

jrmuizel:
This doesn't seem to be working. I don't see any `fog.` probes when searching.

PR #1864, opened 2022-03-10
create backup MVs and update the models and management commands

alekhyamoz:
We have taken a phased approach for deploying the db migrations for the issue #1240 
This PR will create backup MVs which is the 1st step in the deployment process. 
The second migration file will DROP the existing MV and rename the backup to the existing MV.


glam/api/migrations/0023_drop_alter_MV.py
```diff
@@ -0,0 +1,99 @@
+# Generated by Django 3.1.13 on 2022-03-04 01:21
```
edugfilho, line 1: This migration file is dropping the old MV and renaming the backup to make it the effective one. Is this file supposed to be in this PR?

alekhyamoz: Without this file, the tests are failing as discussed in https://mozilla.slack.com/archives/C01QMME608M/p1646873336029739

edugfilho: Got it. Thanks!
I hadn't read that and got confused by your first comment on this PR.

edugfilho:
LGTM
I'd only suggest to change the first comment to reflect that we're doing the whole migration at once, in case this PR get revisited in the future

edugfilho (approved):

Iinh (approved):

glam/api/migrations/0022_create_backup_MV.py
```diff
@@ -0,0 +1,82 @@
...
+for channel in constants.CHANNEL_NAMES.values():
+    sql_backup.extend(
+        [
+            f"CREATE MATERIALIZED VIEW view_glam_desktop_{channel}_aggregation_backup AS SELECT * FROM glam_desktop_{channel}_aggregation;",
+            f"CREATE UNIQUE INDEX ON view_glam_desktop_{channel}_aggregation_backup (id);",
+            f"CREATE INDEX ON view_glam_desktop_{channel}_aggregation_backup (version);",
+            f"CREATE INDEX ON view_glam_desktop_{channel}_aggregation_backup USING HASH (metric);",
+            f"CREATE INDEX ON view_glam_desktop_{channel}_aggregation_backup USING HASH (os);",
+        ]
+    )
```
willkg, line 17: This code runs at import time. You probably want to put this in a function so that it doesn't run every time this file is imported.

Also, it's generally better to write migrations such that they're "set in stone" at the time you wrote/ran them. This one uses `glam.api.constants`. If that changes in the future, then this migration changes.

alekhyamoz: Sorry,  I did not understand this. Can you please elaborate?
The backup MVs will not exist when running the import command. They are dropped in the next migration command. (glam/api/migrations/0023_drop_alter_MV.py)

willkg: I should have looked at other migrations in this app before saying anything. While I tend to write migrations in a way that they're "set in stone" because I feel that's safer and it's bit me in the past, it looks like glam does things differently and that's cool.

Also, if someone squashes the migrations at some point, then all of this is moot.

PR #1866, opened 2022-03-10
Update explore view for probes with only 1 histogram

### LINKED ISSUE #1187: Update explore view for new probes with only 1 histogram
For new probes added it takes GLAM 2 histograms worth of data to show a chart. With only 1 histogram the explore page could be a bit more explanatory or show some data. The usual chart view requires multiple points to show the over-time chart but we could show something else here until more data flows in.

Iinh:
For probes that only have 1 data point, add more information to the Explore view to explain why there's no graph available. Fixes #1187.

Since this was an old error (the bug was reported over a year ago), and there's no new probes added recently, it was hard to find a production example to show what the page looks like with only 1 data point. I had to manually import some test data to reproduce the view locally.

Below is what an explore page looks like when there's only 1 histogram.

Screen Shot 2022-03-09 at 11 52 09 AM

After this change:

Screen Shot 2022-03-10 at 4 08 07 PM

### DIFF
--- a/src/components/explore/ComparisonSummary.svelte
+++ b/src/components/explore/ComparisonSummary.svelte
@@ -138,7 +138,7 @@
       use:tooltipAction={{
         text: `Compares the numeric values of the reference ⭑ to the hovered values ●. ${
           justOne
-            ? `Please note that currently this probe doesn't have enough data to produce a meaningful comparison yet.`
+            ? `Currently we don't have enough data to generate a meaningful comparison yet, please use this chart as a histogram distribution graph.`
             : ''
         }`,
         location: 'top',
--- a/src/components/explore/ComparisonSummary.svelte
+++ b/src/components/explore/ComparisonSummary.svelte
@@ -19,6 +19,7 @@
   export let showRight = true;
   export let showDiff = true;
   export let viewType;
+  export let justOne;
 
   function percentChange(l, r) {
     return viewType === 'proportion' ? r - l : (r - l) / l;
@@ -135,7 +136,11 @@
     Summary
     <span
       use:tooltipAction={{
-        text: 'Compares the numeric values of the reference ⭑ to the hovered values ●',
+        text: `Compares the numeric values of the reference ⭑ to the hovered values ●. ${
+          justOne
+            ? `Please note that currently this probe doesn't have enough data to produce a meaningful comparison yet.`
+            : ''
+        }`,
         location: 'top',
       }}
       class="data-graphic__element-title__icon"><Help size={14} /></span>
--- a/src/components/explore/ProbeExplorer.svelte
+++ b/src/components/explore/ProbeExplorer.svelte
@@ -3,6 +3,7 @@
 
   import { window1D } from '@graph-paper/core/utils/window-functions';
   import ToplineMetrics from './ToplineMetrics.svelte';
+  import GlamErrorShapes from '../errors/GlamErrorShapes.svelte';
 
   import AggregationsOverTimeGraph from './AggregationsOverTimeGraph.svelte';
   import AggregationComparisonGraph from './AggregationComparisonGraph.svelte';
@@ -203,6 +204,23 @@
     justify-items: start;
     margin-bottom: var(--space-4x);
   }
+
+  .data-error-msg__bg {
+    background: radial-gradient(var(--cool-gray-100), var(--cool-gray-100));
+    width: 200px;
+    height: 200px;
+    padding: var(--space-4x);
+    margin: 0 auto;
+    border-radius: 50%;
+    margin-top: var(--space-2x);
+  }
+
+  .detail-title {
+    font-size: var(--text-015);
+    font-weight: 300;
+    color: var(--cool-gray-700);
+    padding: 10px 40px;
+  }
 </style>
 
 <div class="probe-body-overview">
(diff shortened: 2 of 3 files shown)


src/components/explore/ProbeExplorer.svelte
```diff
@@ -215,31 +233,53 @@
...
+        </div>
+        <div>
+          <p class="detail-title">
+            Currently we don't have enough data to generate an over-time graph
+            for this metric. Please refer to the Table View to see the available
+            data, or use the SQL query generator for <a
+              href="https://docs.telemetry.mozilla.org/cookbooks/main_ping_exponential_histograms.html"
+              >further STMO analysis</a
+            >.
+            <br /> <br />
```
edugfilho: I think you can achieve the desired spacing between elements with the `detail-title` css class instead of using `<br />` tags

robhudson:
Yay! I'm glad to see this one get fixed. It was a very confusing state when you can't draw a time series with 1 histogram. This is a nice solution.

src/components/explore/ProbeExplorer.svelte
brizental: Yeap, agreed. Since these are two distinct paragraphs we can can break them into two separate `<p>` tags :)

src/components/explore/ProbeExplorer.svelte
```diff
@@ -215,31 +233,53 @@
...
-      bind:hoverValue
-      {aggregationLevel}
-      on:click={() => {
-        if (hovered.datum) {
-          ref = hovered.datum;
-        }
-      }}>
-      <slot name="additional-plot-elements" />
-    </AggregationsOverTimeGraph>
+  <div style="display: block">
```
brizental: `<div>`s are `display: block` by default. This is possibly redundant.

brizental (commented):
newbie question: What data are the "compare" and "Summary" sections showing? I looked at another GLAM page and the tooltip tells me: "Compares the reference to the hovered value...", if there is no value to hover what are we comparing? Should this PR also include changes to the tooltip texts to explain what the person is seeing?

I have one other super nit: I notice from the screenshot that the notice about not having enough data is not aligned with the title, can we align them?

Iinh:
> newbie question: What data are the "compare" and "Summary" sections showing? I looked at another GLAM page and the tooltip tells me: "Compares the reference to the hovered value...", if there is no value to hover what are we comparing? Should this PR also include changes to the tooltip texts to explain what the person is seeing?

For this specific case, the "compare" view shows the histogram distribution of this metric on that specific date/build ID (the date is shown right above the graph), similar to the "Distribution" view in TMO (but vertically!). Same with "Summary", but for percentile data. 

Agree that with one data point there's really nothing to compare to- I added extra explanation to these tooltips like suggested. Thanks for spotting this!

> I have one other super nit: I notice from the screenshot that the notice about not having enough data is not aligned with the title, can we align them?

They are now aligned  📐 Please see below:

Screen Shot 2022-03-11 at 4 15 28 PM

src/components/explore/ProbeExplorer.svelte
Iinh: thank you both! this is now fixed!

src/components/explore/ComparisonSummary.svelte
```diff
@@ -135,7 +136,11 @@
     Summary
     <span
       use:tooltipAction={{
-        text: 'Compares the numeric values of the reference ⭑ to the hovered values ●',
+        text: `Compares the numeric values of the reference ⭑ to the hovered values ●. ${
+          justOne
+            ? `Please note that currently this probe doesn't have enough data to produce a meaningful comparison yet.`
```
brizental: This is still confusing to me. While we say there is not enough data to produce meaningful comparisons yet, we still show _something_ on the graph. This might just be my lack of GLAM understanding showing, but thought it would be worth pointing out.

brizental (commented):
Hey @Iinh I have one more question :) See above.

With regards the alignment, it looks much better. I think the layout would benefit from a bit of margin between the notice and the "Compare" graph, it looks a bit crammed now. WDYT?

src/components/explore/ComparisonSummary.svelte
Iinh: Because we only have 1 data point now, the graph shows the histogram distribution of that 1 data point. There is still value in showing the graph when it's just 1 data point, even when there's nothing to compared to (hence the "no meaningful comparison" mention), because it still shows the bucket distribution. But I do see how this tooltip text does not make clear of that usage, so I just added a sentence to explain this.

Iinh:
> I think the layout would benefit from a bit of margin between the notice and the "Compare" graph, it looks a bit crammed now. WDYT?

Yeah that makes sense. I think with some margin it looks a bit better now.

Screen Shot 2022-03-14 at 7 53 56 PM

src/components/explore/ComparisonSummary.svelte
brizental: > But I do see how this tooltip text does not make clear of that usage, so I just added a sentence to explain this.

Hey Lihn, did you forget to commit this change? I don't see any changes in the tooltip since last time.

brizental (commented):
It bothers me that there are no tests for this change, but I looked around the repo and it seems there are no front-end tests setup for now, is that correct? Would you kindly file a bug to investigate doing that? I understand it's a big endeavour, but at least having a bug so it's in our minds might be good. 

Looking around the repo I also noticed storybook is setup, but is not being used. It would have made this change much easier to test, because we could control the state more easily. Something to think about for future planning.

I just left one last question, otherwise, for now, this is good to go.

brizental (approved):
Oops, forgot to actually approve.

PR #1871, opened 2022-03-14
Cast version from string to int to fix version 100 data not showing

### LINKED ISSUE #1870: GLAM doesn't show data after `2022-03-08` (version=100)
The data is available in the backend tables, but isn't displayed on the front-end:

Screen Shot 2022-03-14 at 9 46 59 AM

When aggregated by Major Version, GLAM displays data up to 99
Screen Shot 2022-03-14 at 9 47 58 AM

Iinh:
Fixes  #1870.

Version data is stored as string in our database, so after version 100, we need to make sure to cast it from string to int first, or Max() would return version 99 as the latest version, instead of 100, because `"99" > "100"`.

Seeing version 100 data now that version is converted to int. 

Aggregated by build id:

CleanShot 2022-03-14 at 10 58 58@2x

By version:

CleanShot 2022-03-14 at 11 01 35@2x

### DIFF
--- a/glam/tests/test_api.py
+++ b/glam/tests/test_api.py
@@ -321,7 +321,7 @@ class TestDesktopAggregationsApi:
             "process": "parent",
             "total_addressable_market": 999,
             "total_users": 1110,
-            "version": "72",
+            "version": 72,
         }
 
     def test_revision_lookup(self, client):
@@ -359,7 +359,7 @@ class TestDesktopAggregationsApi:
             "process": "parent",
             "total_addressable_market": 999,
             "total_users": 1110,
-            "version": "72",
+            "version": 72,
         }
 
     def test_versions_provided(self, client):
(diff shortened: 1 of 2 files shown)


alekhyamoz (approved):
lgtm

edugfilho (approved):

glam/tests/test_api.py
```diff
@@ -399,7 +399,7 @@ def test_versions_count(self, client):
         data = resp.json()
         assert len(data["response"]) == 4
         versions = sorted([d["version"] for d in data["response"]])
-        assert versions == sorted(["72", "71", "70", "69"])
+        assert versions == sorted([72, 71, 70, 69])
```
robhudson: It would be good to add a test for version 100 to verify in the tests that 100 > 99. You could tweak this one or duplicate it to add another one?

Iinh: Good idea! Let me add that. Thanks Rob!

PR #1876, opened 2022-03-15
Includes sample counts column addition and app version change from string to int

alekhyamoz:
This PR include the following changes: 
1. The GLAM backend tables currently have a [version](url) column as string and casting this as INT is slowing down the front end. For the version 100 to show on front end we need to update the column type for all the backend agg tables. This will require dropping and re-creating the MVs (materialized views)

2. Addition of total_sample column to the agg tables
3. Remove HASH index on [os](url) for desktop MVs and on [app_id](url) for Glean products (needs sufficient performance testing in dev env)


edugfilho:
Wouldn't it be better to have different commits and PRs for each issue?

alekhyamoz:
> Wouldn't it be better to have different commits and PRs for each issue?

The migration file is the same for all the issues. We need to drop and re-create the MVs for each issue. So instead of doing that multiple times this PR alone will solve all 3 issues

glam/api/models.py
```diff
@@ -35,7 +35,7 @@ class AbstractGleanAggregation(models.Model):
     # Dimensions.
     app_id = models.CharField(max_length=100)
     channel = models.CharField(max_length=100)
-    version = models.CharField(max_length=100)
+    version = models.IntegerField()
```
edugfilho: it was being casted as a BigInteger before, but I guess Integer should do :p

edugfilho (approved):

robhudson (approved):
The code looks okay to me. I hope it helps with the materialized view processing.

Iinh (approved):

edugfilho (approved):

cmharlow:
The DB migrations have been applied successfully in dev. Merging so the automated application deploy associated with the db migrations can now deploy.

glam/api/views.py
```diff
@@ -98,11 +98,11 @@ def get_firefox_aggregations(request, **kwargs):
     # Whether to pull aggregations by version or build_id.
     if aggregation_level == "version":
         dimensions.append(Q(build_id="*"))
-        counts = _get_firefox_counts(channel, os, versions, by_build=False)
+        #counts = _get_firefox_counts(channel, os, versions, by_build=False)
```
edugfilho, line 101: GLAM has stopped using separate counts in favor of `total_users`column embedded in the probe's aggregates

PR #1892, opened 2022-03-23
Get the correct expiry status for Fenix probes

Iinh:
The Glean Dictionary search API returns `null` for probes that don't have an expiry date, i.e. they never expire. However we're not including this condition in the expiry check for front-end, which marks active probe as expired.

For example, this probe is still active, but on dev it's marked as expired: https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/addons_has_enabled_addons/explore?activeBuckets=%5B%22never%22%2C%22always%22%2C%22sometimes%22%5D&aggType=avg&currentPage=1

It's because the API returns the expiry value as `null`: https://dictionary.telemetry.mozilla.org/api/v1/metrics_search_fenix?search=addons.has_enabled_addons

This PR fixes it, and the probe now has the correct expiry status.

CleanShot 2022-03-23 at 15 44 23@2x

### DIFF
--- a/src/routing/pages/probe/FenixDetails.svelte
+++ b/src/routing/pages/probe/FenixDetails.svelte
@@ -20,11 +20,12 @@
   const mdY = timeFormat('%b %d, %Y');
   const toNiceDate = (dt) => mdY(parseYMD(dt));
 
+  const willNeverExpire = (expiry) =>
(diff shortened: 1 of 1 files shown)


edugfilho (approved):
LGTM.
thanks for adding the api call to the description!

PR #1895, opened 2022-03-24
fenix model version lookup gets top 3 distinct istead of top version & subtract 3

cmharlow:
Jira ticket: https://mozilla-hub.atlassian.net/browse/DSRE-632

Note: if you prefer to stick with focusing on the ETL data validation & filtering alone, that's cool. Just offering another option to update the app to be smarter (or maybe just more explicit - a subtract 3 approach has a baked in assumption of versions that may not always be true).

What this PR does:
* the Fenix probe lookups were failing bc there was a non-firefox desktop version present that was used as the max version
* the view would then calculate the query with a version filter that was the max version minus the number of versions to retrieve (e.g. 2021, 2020, 2019) - or whatever 
* this caused Fenix probe lookups to fail as the query returned no docs
* this PR instead has the get_glean_aggregations method calcuate the versions to query as the top (number of version desired, defaults to  3) distinct versions in the given view instead of the top version and substract.
* a possible con of this PR is the query change could be a performance impact - replacing, for Fenix products only, the first query below (plus a subtraction) to the second query below, which we know gets us versions we want:

old query:
```
SELECT MAX("view_glam_fenix_aggregation"."version") AS "version__max" FROM "view_glam_fenix_aggregation

# sql response
2015819723

# python versions list calculation
['2015819723', '2015819722', '2015819721']
```

new query:
```
SELECT DISTINCT ON ("view_glam_fenix_aggregation"."version") "view_glam_fenix_aggregation"."version" FROM "view_glam_fenix_aggregation" ORDER BY "view_glam_fenix_aggregation"."version" DESC LIMIT 3;

# sql response
[2015819723, 100, 99]

# python versions list calculation
[2015819723, 100, 99]
```

### DIFF
--- a/glam/api/views.py
+++ b/glam/api/views.py
@@ -195,16 +195,19 @@ def get_glean_aggregations(request, **kwargs):
     product = kwargs.get("product")
 
     num_versions = kwargs.get("versions", 3)
+    versions = list()
     try:
         max_version = int(model.objects.aggregate(Max("version"))["version__max"])
+        versions = list(model.objects\
+            .order_by('-version')\
+            .values_list('version', flat=True)\
+            .distinct('version')[:num_versions])
     except (ValueError, KeyError):
         raise ValidationError("Query version cannot be determined")
     except TypeError:
-        # This happens when `version_max` is NULL and cannot be converted to an int,
+        # This happens when `version` is NULL,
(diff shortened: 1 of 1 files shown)


akkomar:
Superseded by https://github.com/mozilla/glam/pull/2129

PR #1909, opened 2022-04-05
Show sample count on Explore page

### LINKED ISSUE #1851: Add sample counts to GLAM UI
Contingent to #1240. Once the sample counts data is available in Postgres tables, the sample counts need to be displayed front-end.

Iinh:
For now this PR focuses on surfacing sample count data for desktop probes. Sample counts for Glean metrics (Fenix and FOG) will be added later (note that sample count data is already available for both legacy and Glean metrics in GLAM db, thanks to #1853).

Main UI changes:

CleanShot 2022-04-04 at 20 37 25@2x
1. Add a new dropdown that allows toggling between user count, sample count, or both
2. Add sample count data above the graph, next to the user count information
3. Sample count volume graph is below the aggregation graph

Here's what the UI looks like when "both" option is selected:

CleanShot 2022-04-04 at 20 39 31@2x

Fixes #1851.

### DIFF
--- a/src/components/explore/ClientVolumeOverTimeGraph.svelte
+++ b/src/components/explore/ClientVolumeOverTimeGraph.svelte
@@ -29,6 +29,14 @@
   export let hoverValue = {};
 </script>
 
+<style>
+  .count-view {
+    font-weight: 300;
+    font-size: 0.7em;
+    cursor: pointer;
+  }
+</style>
+
 <div>
   <ChartTitle
     {description}
@@ -36,7 +44,7 @@
     right={totalClientsGraph.right}>
     {title}
     <a
-      style="font-weight: 300; font-size: 0.7em; cursor: pointer;"
+      class="count-view"
       on:click={() => store.setField('countView', 'samples')}
       >View Sample Count</a>
   </ChartTitle>
--- a/src/components/explore/ClientVolumeOverTimeGraph.svelte
+++ b/src/components/explore/ClientVolumeOverTimeGraph.svelte
@@ -1,6 +1,7 @@
 <script>
   import { Axis } from '@graph-paper/guides';
   import { Line } from '@graph-paper/elements';
+  import { store } from '../../state/store';
 
   import Tweenable from '../Tweenable.svelte';
   import DataGraphic from '../datagraphic/DataGraphic.svelte';
@@ -34,6 +35,10 @@
     left={totalClientsGraph.left}
     right={totalClientsGraph.right}>
     {title}
+    <a
+      style="font-weight: 300; font-size: 0.7em; cursor: pointer;"
+      on:click={() => store.setField('countView', 'samples')}
+      >View Sample Count</a>
   </ChartTitle>
   <DataGraphic
     yType="linear"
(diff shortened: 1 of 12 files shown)


Iinh:
After discussion we decided to make the toggle just a link right above the volume chart, instead of a top-level dropdown menu. Here's what it looks now to switch between Client and Sample:


Also, this PR now add sample count data to _all products_, not just desktop.

src/components/explore/ClientVolumeOverTimeGraph.svelte
```diff
@@ -34,6 +35,10 @@
     left={totalClientsGraph.left}
     right={totalClientsGraph.right}>
     {title}
+    <a
+      style="font-weight: 300; font-size: 0.7em; cursor: pointer;"
```
edugfilho: Would it be better to have this in a class?

src/components/explore/ProbeExplorer.svelte
```diff
@@ -150,9 +155,14 @@
   // client counts.
   const MULT = 1.1;
   $: clientCountsData = clientCounts(data);
+  $: sampleCountsData = sampleCounts(data);
+  $: yValsSample = sampleCountsData.map((d) => d.totalSample);
   $: yVals = clientCountsData.map((d) => d.totalClients);
-  $: yMax = Math.max(50, Math.max(...yVals));
-  $: yClientsDomain = [0, yMax * MULT];
+  $: yMaxClient = Math.max(50, Math.max(...yVals));
```
edugfilho: Can we add a `const` for this `50` instead of leaving it as a [Magic Number](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)?

edugfilho, line 161: Can we add a `const` for this `50` instead of leaving it as a [Magic Number](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)?

src/components/explore/SampleCountOverTimeGraph.svelte
```diff
@@ -0,0 +1,141 @@
...
+</script>
+
+<div>
+  <ChartTitle
+    {description}
+    left={totalClientsGraph.left}
+    right={totalClientsGraph.right}>
+    {title}
+    <a
+      style="font-weight: 300; font-size: 0.7em; cursor: pointer;"
```
edugfilho: Would it be better to have this in a class?

edugfilho, line 39: Would it be better to have this in a class?

edugfilho (commented):
Sorry for the delay on reviewing this!

src/components/explore/ProbeExplorer.svelte
edugfilho: this comment is a dupe! IDK what happened.

src/components/explore/SampleCountOverTimeGraph.svelte
edugfilho: this comment is a dupe! IDK what happened.

src/components/explore/ClientVolumeOverTimeGraph.svelte
Iinh: Makes sense. Class added.

edugfilho (approved):
great stuff! lgtm

PR #1933, opened 2022-04-14
Fix FOG bugs: Looker links, release marker, ping types

Iinh:
This PR fixes the bugs that I found while testing FOG page on dev. Details are in the comment below. 

There is still one outstanding bug: we're not getting the correct version data. For some reason, we're getting `1024` as the version number (instead of standard Firefox versions 98, 99, 100) for all Glean probes now. This is most likely a bug in our ETL.

CleanShot 2022-04-14 at 15 32 47@2x

Database check:

```
postgres=# select version from glam_fog_aggregation group by version;
 version 
---------
    1024
(1 row)
```

### DIFF
--- a/src/components/ReleaseVersionMarkers.svelte
+++ b/src/components/ReleaseVersionMarkers.svelte
@@ -10,7 +10,7 @@
   const xScale = getContext('xScale');
 
   let markers = [];
-  switch ($store.product) {
+  switch ($store.searchProduct) {
     case 'firefox':
       firefoxVersionMarkers.subscribe((m) => {
         markers = m;
@@ -26,7 +26,7 @@
   }
 </script>
 
-{#if ['firefox', 'fenix'].includes($store.product)}
+{#if ['firefox', 'fenix'].includes($store.searchProduct)}
   <g class="firefox-release-version-markers">
     {#if markers && markers.length}
       {#each markers.filter((d) => d.date !== undefined && d.date >= $xScale.domain()[0] && d.date <= $xScale.domain()[1]) as { label, date }, i (date)}
--- a/src/config/glean.js
+++ b/src/config/glean.js
@@ -37,8 +37,12 @@ export const FIREFOX_ON_GLEAN = {
     ping_type: {
       title: 'Ping Type',
       key: 'ping_type',
-      values: [{ key: '*', label: 'All' }],
-      defaultValue: '*',
+      values: [
+        { key: '*', label: 'All' },
+        { key: 'metrics', label: 'Metrics' },
+        { key: 'baseline', label: 'Baseline' },
+      ],
+      defaultValue: 'metrics',
       isValidKey(key, probe) {
         return key === '*' ? true : probe.send_in_pings.includes(key);
       },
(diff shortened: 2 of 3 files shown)


src/components/ReleaseVersionMarkers.svelte
```diff
@@ -10,7 +10,7 @@
   const xScale = getContext('xScale');
 
   let markers = [];
-  switch ($store.product) {
+  switch ($store.searchProduct) {
```
Iinh, line 13: Currently getting this error in the console for all FOG metrics:

CleanShot 2022-04-14 at 15 29 32@2x

We should use `searchProduct` instead of `product` to be inclusive of FOG.

src/config/glean.js
```diff
@@ -37,8 +37,12 @@ export const FIREFOX_ON_GLEAN = {
...
       title: 'Ping Type',
       key: 'ping_type',
-      values: [{ key: '*', label: 'All' }],
-      defaultValue: '*',
+      values: [
+        { key: '*', label: 'All' },
+        { key: 'metrics', label: 'Metrics' },
+        { key: 'baseline', label: 'Baseline' },
+      ],
+      defaultValue: 'metrics',
```
Iinh, line 45: I assumed that FOG metrics have just one ping type like Fenix metrics -- `metrics` -- but this is not right. I ran a query against the database to check how many ping types are being sent to the ETL now and got 3 total.

```
postgres=# select ping_type from glam_fog_aggregation group by ping_type;
 ping_type 
-----------
 *
 metrics
 baseline
(3 rows)
```

src/routing/pages/probe/GleanDetails.svelte
```diff
@@ -185,7 +190,9 @@
           {@html marked($store.probe.description)}
           <a
             class="more-info-link"
-            href={`https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/${$store.probeName}`}
+            href={`https://dictionary.telemetry.mozilla.org/apps/${
+              GLEAN_DICTIONARY_PRODUCT_IDS[$store.product]
+            }/metrics/${$store.probeName}`}
```
Iinh, line 195: Get the current Glean Dictionary link based on the product.

src/routing/pages/probe/GleanDetails.svelte
```diff
@@ -267,9 +277,9 @@
         Export to JSON
       </button>
       <LookerLink
-        product="fenix"
+        product="glean"
         variants={$store.probe.variants}
-        sendInPings={$store.probe.send_in_pings}
+        sendInPings={$store.productDimensions.ping_type}
```
Iinh, line 282: FOG page LookerLink area is greyed out:

CleanShot 2022-04-14 at 15 17 27@2x

which indicates that there's no Looker available for FOG probes, which is not true. Because FOG now have different ping types, we need to change how we get the ping information for LookerLink component.

CleanShot 2022-04-14 at 15 13 23@2x

edugfilho (approved):
lgtm. thanks for those fixes!

PR #1934, opened 2022-04-14
Get the correct release marker component path for sample count chart

Iinh:
Oops forgot to change to the correct ReleaseMarker component in the last PR.


Iinh:
Merging this now due to this being just the wrong path name.

PR #1966, opened 2022-05-03
Remove process filter in SQL query when there is no process

### LINKED ISSUE #1955: View SQL Query filtering for `process=""` for probes not found
When a probe is not found, the [The SQL window](https://github.com/mozilla/glam/blob/28e8fa00a302d4ec92ece77a51c9206265440ec7/src/components/SqlModal.svelte#L45) populates the `process` filter with an empty string, which is not good because that filter prevents us from finding anything for that probe on Redash.

Granted, most of the time when a probe is not found it is not going to be on Redash either, but in cases like #1954 it is useful to _not_ have such filter, so we can use the query to double check whether the data actually exists.

Iinh:
Fixes #1955.

### DIFF
--- a/src/components/SqlModal.svelte
+++ b/src/components/SqlModal.svelte
@@ -39,12 +39,15 @@
       $store.productDimensions.aggregationLevel === 'build_id'
         ? 'app_build_id IS NOT NULL'
         : 'app_build_id IS NULL';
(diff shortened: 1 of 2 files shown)


edugfilho (approved):
lgtm
Makes much more sense now plus It will help us debug things.
Thanks for working on this!

PR #1991, opened 2022-05-23
Fix broken Looker links

Iinh:
When Looker data is not available, don't link to it. This prevents broken pages due to unavailable Looker links.

For example: https://dev.glam.nonprod.dataops.mozgcp.net/fog/probe/glean_baseline_duration/explore?aggType=avg&app_id=release&currentPage=1&ping_type=* 

is broken and the error message is:

`Uncaught (in promise) TypeError: can't access property "looker", r.etl.ping_data[s] is undefined`

CleanShot 2022-05-23 at 18 25 01@2x

### DIFF
--- a/src/components/LookerLink.svelte
+++ b/src/components/LookerLink.svelte
@@ -17,6 +17,7 @@
 
     lookerURL =
       variant &&
+      variant.etl.ping_data[sendInPings] &&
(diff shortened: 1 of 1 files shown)


edugfilho (approved):
lgtm. Thanks for fixing this!

PR #2011, opened 2022-06-07
Make header items more responsive

Iinh:
This PR fixes 2 header problems when the viewport is smaller than 1045px.

1) GLAM logo is hidden behind the search bar

CleanShot 2022-06-07 at 12 43 02@2x

2) Overflown items at the end

CleanShot 2022-06-07 at 12 41 56@2x
 
This is how the app looks with this change: the logo is always visible, and no more overflow items at the end.

### DIFF
--- a/public/static/global.css
+++ b/public/static/global.css
@@ -408,10 +408,11 @@ button:focus {
 .app {
   display: grid;
   min-height: 100vh;
-  max-width: var(--app-width);
+  min-width: var(--app-width);
   grid-template-rows: max-content auto;
   margin-left: auto;
   margin-right: auto;
+  padding: 0 10px;
 }
 
 .app--toolbar {
--- a/src/components/icons/GLAMLogo.svelte
+++ b/src/components/icons/GLAMLogo.svelte
@@ -2,7 +2,7 @@
   import { onMount } from 'svelte';
   import { fly } from 'svelte/transition';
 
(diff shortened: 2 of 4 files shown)


public/static/global.css
```diff
@@ -408,10 +408,11 @@ button:focus {
 .app {
   display: grid;
   min-height: 100vh;
-  max-width: var(--app-width);
+  min-width: var(--app-width);
```
Iinh, line 411: we should use `min-width` instead of `max-width` here to make sure that the app width always takes (almost, minus the padding) as much space as the allocated `app-width` values.

src/components/icons/GLAMLogo.svelte
```diff
@@ -2,14 +2,17 @@
...
   let mounted = false;
   onMount(() => {
     mounted = true;
   });
   let y = 50;
   let duration = 400;
   let delay = 200;
+
+  let innerWidth = { window };
+  $: size = innerWidth < 1045 ? 22 : 24;
```
edugfilho: I'd suggest a comment or variable to clarify where 1045 comes from (same for GLAMText)

Iinh: good call!

edugfilho (approved):
looks awesome!
thank you

PR #2012, opened 2022-06-07
Add GLEAN visualizer to FOG metric names

Iinh:
Make it more clear, besides the url signal, whether a desktop metric is a Glean metric.

CleanShot 2022-06-07 at 15 53 02@2x


edugfilho (approved):
LGTM
that's a nice addition!

PR #2013, opened 2022-06-07
Fix overflow sample count text

Iinh:
Before

CleanShot 2022-06-07 at 16 19 04@2x

After (no more overflow text)

CleanShot 2022-06-07 at 16 18 38@2x


edugfilho (approved):
lgtm!

PR #2014, opened 2022-06-07
Update CircleCI Docker image

Iinh:
The latest merged PRs were not automatically deployed to dev due to CI unable to find the Docker image. This is because we have an outdated CircleCI orb. Updating to v2.1.1 seems to fix the issue.

CleanShot 2022-06-07 at 16 54 12@2x

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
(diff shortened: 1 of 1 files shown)


edugfilho (approved):
awesome work. Thanks for fixing this!

PR #2030, opened 2022-06-23
sample reworking of CircleCI, Dockerfile to work with Dockerflow

### LINKED ISSUE #1887: Update Dockerflow Application Endpoints for Glam App
According to the [Dockerflow specification](https://github.com/mozilla-services/Dockerflow) (a SRE practice that Glam somewhat follows currently), the Glam application should support 3 endpoints:

1. https://glam.telemetry.mozilla.org/__version__ (this currently fails missing a file)
2. https://glam.telemetry.mozilla.org/__heartbeat__ (this works but the check itself shows warning - we should understand why)
3. https://glam.telemetry.mozilla.org/__lbheartbeat__ (this doesn't work currently)

If SRE gets to this first, we'll do it - but noting this here for the Glam Dev team to call out the version endpoint in particular. This should be a pretty straight forward addition to the CI & Dockerfile (see the example in the linked repo above). 

Linked SRE ticket: https://mozilla-hub.atlassian.net/browse/DSRE-621

cmharlow:
Ticket: #1887 

What this PR does:
* reworks Dockerfile, docker-compose, circleci configs to leverage dockerflow expectations (https://github.com/mozilla-services/Dockerflow - basically, create a version.json file as part of Docker image build & ship with deploy; also run tests in Docker image instead of CI OS)
* adds a docker-compose file for running tests on CircleCI OS (limitation of running tests on our newly built docker image while connecting to a database docker container - we need docker-compose to tell circleci to put these containers on the same docker network, cannot do it ourselves)
* removes bin folder, since unused / artifacts of old build process from few years ago


Iinh (approved):
Looks great 🎉 Thanks for working on this!

PR #2038, opened 2022-06-28
Make sure build_id always has a value in Distribution sql

Iinh:
and also reformat the SQL template. This should incorporate most of @rafrombrc's Slack feedback.

Screen Shot 2022-06-28 at 4 08 37 PM

### DIFF
--- a/src/stringTemplates/desktop-distribution.tpl
+++ b/src/stringTemplates/desktop-distribution.tpl
@@ -1,7 +1,3 @@
--- Please ensure that you have selected a BUILD ID (by clicking
--- on the desired reference point) before proceeding with this
--- template in STMO.
-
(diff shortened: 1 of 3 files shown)


src/stringTemplates/desktop-distribution.tpl
```diff
@@ -1,7 +1,3 @@
--- Please ensure that you have selected a BUILD ID (by clicking
--- on the desired reference point) before proceeding with this
--- template in STMO.
-
 -- For more info on the tables referenced in this SQL see:
 -- https://docs.telemetry.mozilla.org/datasets/main_ping_tables.html
```
Iinh, line 2: I'll open a separate PR in data docs to update instruction on how to create distribution view.

edugfilho (approved):

Issue #1888, opened 2022-03-22
Repair / Annotate Models not managed by Django To Handle Migration Warnings


cmharlow:
During django migrations, the following warning appears:

```
Your models have changes that are not yet reflected in a migration, and so won't be applied.
Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

The idea during triage of a different error is this is related to FOGAggregationView (which is not managed by Django).

This is a super low priority request, but it would be good to confirm this View is indeed the source of this warning, and to silent this warning (perhaps by updating the model with a 'managed=false' addition to the model's meta options).

alekhyamoz:
Need set to 'abstract=True'

edugfilho:
I wonder  why [FenixAggregationView](https://github.com/mozilla/glam/blob/17b5e1576f9c083610c30caa319b57596262d06a/glam/api/models.py#L79), which is similar, doesn't behave the same way.

edugfilho:
Fixed in https://github.com/mozilla/glam/pull/2084

edugfilho:
Reopening this since the previous fix had to be rolled back. Adding `abstract=True` to the model, which is not an abstract one, prevents it from being instantiated hence throwing an exception at such attempt.

PR #2084, opened 2022-08-09
#1888 fix migration warning on FOGAggregationView

edugfilho:
Added `abstract = True` to FOGAggregationView and that removed the warning.

Tested locally that a full first-time migration still works after the change.

I'm still wondering why [FenixAggregationView](https://github.com/mozilla/glam/blob/17b5e1576f9c083610c30caa319b57596262d06a/glam/api/models.py#L79), which is similar, doesn't behave the same way.

### DIFF
--- a/glam/api/models.py
+++ b/glam/api/models.py
@@ -107,6 +107,7 @@ class FOGAggregation(AbstractGleanAggregation):
 class FOGAggregationView(AbstractGleanAggregation):
     class Meta:
         managed = False
+        abstract = True
         db_table = "view_glam_fog_aggregation"
 
 


Iinh (approved):
Looks good to me. Thanks!

chutten:
This change is suspected to be the cause of FOG metrics all 404'ing: https://mozilla.slack.com/archives/CB1EQ437S/p1662413232103959?thread_ts=1662209260.101089&cid=CB1EQ437S @edugfilho do you have any ideas?

edugfilho:
I just tested without this change locally and am still getting 404s, without the error 
`AttributeError: type object 'FOGAggregationView' has no attribute 'objects'`
@Iinh mentioned in the slack thread. 
Still investigating

edugfilho:
Sorry, my local environment was messed up. After a fresh setup I confirmed that this really broke things.
The PR above to fixes it.

PR #2091, opened 2022-08-16
Make violin plot match the ref/hov order

Iinh:
This change includes:
- Match the violin plot direction to hovered/ref order. 
- Match the sample and client count compare graph to hovered/ref order as well.
- Change the label from "Hov"/"Ref" to "Left"/"Right" to better reflect the new UI


edugfilho (approved):
Awesome work!

PR #2129, opened 2022-09-07
Fetch top 3 versions per probe before querying for data

akkomar:
This fixes a problem that we've been seeing from time that was caused by unofficial builds with high version numbers leaking into ETL aggregates.
When probes from such build (say `999` when the latest official build is `106`) were in the database, GLAM backend was filtering for versions (`999`, `998`, `997`) when fetching probe data which in most cases returned empty results (because usually these unofficial versions did not contribute much data).
This PR makes backend to first fetch 3 latest versions for the requested probe that exist in the aggregate table which should fix the problem.

This is inspired by https://github.com/mozilla/glam/pull/1895 by @cmharlow, the only difference being the metric filtering that is added here.

### DIFF
--- a/glam/api/views.py
+++ b/glam/api/views.py
@@ -196,21 +196,24 @@ def get_glean_aggregations(request, **kwargs):
     }
     model = MODEL_MAP[kwargs.get("product")]
     product = kwargs.get("product")
+    probe = kwargs["probe"]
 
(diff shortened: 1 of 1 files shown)


edugfilho (approved):
lgtm. thank you!

PR #2138, opened 2022-09-13
Get the latest build versions from Product Details API

### LINKED ISSUE #1901: Revisit/Rewrite view versions lookup to ignore the inaccurate versions.
The current implementation logic relies on the assumption that the version ids are increments for 1. 
For Firefox: https://github.com/mozilla/glam/blob/a6f9c81722f6d2e4244945daebba19087ea8d5fe/glam/api/views.py#L75
For Fenix: https://github.com/mozilla/glam/blob/a6f9c81722f6d2e4244945daebba19087ea8d5fe/glam/api/views.py#L199 

However, we ran into an issue where bad app_versions are being sent in the pings and eventually pushed to Postgres tables. 
(Eg: 2015819723) resulting in the [code](https://github.com/mozilla/glam/blob/a6f9c81722f6d2e4244945daebba19087ea8d5fe/glam/api/views.py#L198) returning a set of versions that result in no queries. 
Also, the inbuilt logic might not apply to other to-be onboarded products. 

Issue tracked: #1893 

Need to revisit the version lookup logic to avoid the above-mentioned issues
charlow from DSRE has created a [PR](https://github.com/mozilla/glam/pull/1895)

Iinh:
Fixes #1901.

### DIFF
--- a/glam/api/views.py
+++ b/glam/api/views.py
@@ -2,9 +2,10 @@ import os
 
 import dateutil.parser
 import orjson
+import requests
 from django.conf import settings
 from django.core.cache import caches
-from django.db.models import Max, Q, Value, Count
+from django.db.models import Q, Value, Count
 from django.db.models.functions import Concat
 from rest_framework.decorators import api_view
 from rest_framework.exceptions import NotFound, ValidationError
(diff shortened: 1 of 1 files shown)


glam/api/views.py
```diff
@@ -26,6 +27,25 @@
     UsageInstrumentation,
 )
 
+FIREFOX_DESKTOP_PRODUCT_DETAILS = requests.get(
+    "https://product-details.mozilla.org/1.0/firefox_versions.json"
```
akkomar: When and how often will this be called in application lifecycle? During startup of the process?
If so: is there a risk we might get stale version numbers here?

akkomar (commented):
Just a quick first look with a question inline.

Would it be hard to add a test that would cover this change? I'm thinking something that would also document our expectation for the Product Details API response.

glam/api/views.py
Iinh: This is a great point. Let me move this into the get aggregations method to avoid the risk of stale version numbers.

glam/api/views.py
```diff
@@ -80,6 +82,8 @@ def get_firefox_aggregations(request, **kwargs):
     product = "firefox"
     channel = kwargs.get("channel")
     model_key = f"{product}-{channel}"
+    product_details = requests.get(PRODUCT_DETAILS_URLS[product]).json()
```
akkomar, line 85: It seems that now we call this API on each GET request. Would it visibly slow down the application? I think caching these values might make sense (would need to be periodically refreshed - either here or asynchronously). What do you think?

PR #2139, opened 2022-09-13
Make it more clear that View SQL query is not available for FOG yet

Iinh:
Recently we got a user asking why it doesn't return anything when they clicked on View SQL query for FOG metric. Even though we're working on providing it, until we have it available let's add some UI signal that it's not available yet.


edugfilho (approved):
lgtm. Thanks!

PR #2140, opened 2022-09-13
Fix big number formatting for client volume graph

Iinh:
Upon checking on dev I realized that we should also add the number formatting to the compare chart as well. 

On dev currently: https://dev.glam.nonprod.dataops.mozgcp.net/firefox/probe/loaded_tab_count/explore?channel=release&currentPage=1&process=parent
CleanShot 2022-09-13 at 12 12 44@2x

After:
CleanShot 2022-09-13 at 12 17 01@2x


edugfilho (approved):
good catch

PR #2146, opened 2022-09-20
Don't use default yDomain scale for timing distributions

Iinh:
Because we want the line chart scale to match the violin plot's when the user first lands on the page, [we set the y-domain scale to be the default range if there are 5 (default) selected percentiles](https://github.com/mozilla/glam/blob/877daf4ababa56c6e984654bb9f9fe6e758696bc/src/components/explore/AggregationsOverTimeGraph.svelte#L203). However, there are cases when this doesn't work well, especially with timing distribution metrics, as recently reported in our Slack channel. 

This PR makes sure the y-domain for timing distributions is always within range. Let's discuss more in the group meeting to see whether or not we should remove this condition for all metrics.


edugfilho (approved):
lgtm!

PR #2147, opened 2022-09-20
Add instruction on how to update Python dependencies

Iinh:


edugfilho (approved):
thanks for doing that!

Issue #2145, opened 2022-09-15
GLAM visualizations are missing data: should have more than 3 months


mcomella:
**Describe the bug**
For at least two fenix `timing_distribution` probes, [`perf.awesomebar.bookmark_suggestions`](https://glam.telemetry.mozilla.org/fenix/probe/perf_awesomebar_bookmark_suggestions/explore?aggType=avg&currentPage=1&timeHorizon=ALL&visiblePercentiles=%5B95%2C75%2C50%2C25%2C5%2C1%5D) and [` perf.startup.cold_main_app_to_first_frame`](https://glam.telemetry.mozilla.org/fenix/probe/perf_startup_cold_main_app_to_first_frame/explore?aggType=avg&currentPage=1&timeHorizon=ALL&visiblePercentiles=%5B95%2C75%2C50%2C25%5D), GLAM only shows data going back to late June but queries in [STMO](https://sql.telemetry.mozilla.org/queries/87694/source) and [Looker](https://mozilla.cloud.looker.com/explore/fenix/metrics?qid=EULVUWcqucqxIQsDZNEt93&toggle=fil,vis) show that we have data going back to at least January. GLAM should be able to show that data too.

**To Reproduce**
1. Click the GLAM and STMO/Looker links above to compare what data is available.

**Expected behavior**
GLAM has data going back to at least January.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Firefox
 - Version 106.0a1 2022-09-14

**Smartphone (please complete the following information):**
N/A

**Additional context**

edugfilho:
Looks like at each run the ETL is replacing the Glam sql fenix tables with the last 3 versions of data. Will keep investigating

edugfilho:
Scratch that. Inspecting the sql fenix tables I can see data for `perf_awesomebar_bookmark_suggestions` all the way to `2021040717`. I also confirmed the backend service doesn't return the older data. I'll see why.

edugfilho:
I noticed the Glam front-end does not specify the amount of versions when it requests Glean probe data, which makes the back-end default to 3, whereas Firefox desktop (legacy) asks for 10 versions.

I added the `versions: 10` parameter to the Glean code [in this branch](https://github.com/mozilla/glam/tree/glean-req-10-versions) but that only gives about an extra month in the chart. I'll work more on this next week.

Also paging @Iinh because there's a good chance she'll be able to tell what's going on.

PR #2193, opened 2022-10-24
Request up to 20 versions of data for all products

Iinh:
This addresses 2 issues:
- fix bug #2145 (thanks @edugfilho for the investigation, I can confirm that adding `versions: 10` to the request will return 10 versions of data for Glean products)
- since we have more than 10 versions of data in our cloud database, I think it's more useful to request more -- up to 20 versions to make available under the "All" tab

CleanShot 2022-10-25 at 01 40 54@2x


edugfilho:
LGTM. Thank you for confirming and pushing this fix!

edugfilho (approved):

PR #2203, opened 2022-11-03
Remove workaround for probeinfo always returning gzip

edugfilho:
The endpoint no longer returns gzip content


alekhyamoz (approved):
lgtm

PR #2216, opened 2022-11-08
Show Glean expiry version number along with expiry date

Iinh:
As of currently, Glean apps allow expiry information to be either a version number or a date, if it's not `never` or `null`. This PR handles displaying this information on the front-end.

The example below shows how we handle the expiry date if it's a version number.

CleanShot 2022-11-08 at 07 18 16@2x

Let's only merge this once https://github.com/mozilla/glean-dictionary/pull/1499 is merged and deployed to prod.


edugfilho (approved):
LGTM; thanks for working on this!

PR #2223, opened 2022-11-14
Validate metric type before fetching aggregation data (fixes #2199)

### LINKED ISSUE #2199: Events are not supported but GLAM doesn't explicitly informs it
GLAM currently does not support event metrics. However they're still searchable, and once an event is selected a `404: No data found for the selected dimensions.` message is displayed, which might lead users to think GLAM supports events but the selected one doesn't have data, which is not true.

Being explicit about not supporting event metrics would avoid confusion.
Suggestions by @chutten:
1) Hide events from the search dropdown 
2) Provide a custom "metrics of this type aren't yet supported" message

Iinh:
For each product, we have a list of metrics we're currently supporting. If a metric type is not in this list, we throw an error message. However, because the type validation step is after fetching data, if there's no aggregation data for a type, it throws an ambiguous 404 error instead of "metric not supported", which causes problems like #2199.

This PR fixes the problem by adding the type check before we fetch aggregation data for Fenix and FOG. In the error message, we also link out to the dictionary to direct users to more resources.

Events should now throw the correct error message. Example:

CleanShot 2022-11-14 at 17 21 34@2x

### DIFF
--- a/src/components/errors/DataErrorGeneric.svelte
+++ b/src/components/errors/DataErrorGeneric.svelte
@@ -1,8 +1,15 @@
 <script>
   import GlamErrorShapes from './GlamErrorShapes.svelte';
+  import { store } from '../../state/store';
 
   export let reason;
   export let moreInformation;
+
+  const DICTIONARY_LINKS = {
+    firefox: `https://probes.telemetry.mozilla.org/?view=detail&probeId=${$store.probe.id}`,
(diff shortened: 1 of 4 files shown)


edugfilho (approved):
LGTM. 
I think we should reuse the `fetchData` function because there's repetition there, but I understand it's the way it was before.
Thank you

PR #2225, opened 2022-11-15
Handle Expired labels for version-based expiry

Iinh:
For version-based expiry we're incorrectly labeling them as "expired" (eg [this metric](https://dev.glam.nonprod.dataops.mozgcp.net/fenix/probe/perf_awesomebar_bookmark_suggestions/explore?aggType=avg&currentPage=1)). I fixed this on the Glean Dictionary already but still need to do this for GLAM.

### DIFF
--- a/src/routing/pages/probe/GleanDetails.svelte
+++ b/src/routing/pages/probe/GleanDetails.svelte
@@ -34,7 +34,7 @@
     // if expires is a number, ie a version number, then compare it
     // to the latest release version (only applied to Fx products)
     if (!Number.isNaN(item.expires) && item.latest_fx_release_version) {
-      return Number(item.latest_fx_release_version) > item.expires;
+      return Number(item.latest_fx_release_version) >= item.expires;
     }
     return new Date() > new Date(item.expires);
   };
--- a/src/routing/pages/probe/GleanDetails.svelte
+++ b/src/routing/pages/probe/GleanDetails.svelte
@@ -31,6 +31,11 @@
   // taken from glean dictionary
   const isExpired = (item) => {
     if (willNeverExpire(item.expires)) return false;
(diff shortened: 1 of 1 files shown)


src/routing/pages/probe/GleanDetails.svelte
```diff
@@ -31,6 +31,11 @@
   // taken from glean dictionary
   const isExpired = (item) => {
     if (willNeverExpire(item.expires)) return false;
+    // if expires is a number, ie a version number, then compare it
+    // to the latest release version (only applied to Fx products)
+    if (!Number.isNaN(item.expires) && item.latest_fx_release_version) {
+      return Number(item.latest_fx_release_version) > item.expires;
```
edugfilho: A couple things here I don't understand:
* Does `item.latest_fx_release_version` keep increasing after the metric expires?
* When `expires` == `latest_fx_release_version` doesn't that mean the item is expired?

Iinh: > Does item.latest_fx_release_version keep increasing after the metric expires?

`item.latest_fx_release_version` is the latest release version. Since we get this metadata from the Glean Dictionary (and the dictionary [fetches it from the Product Details API](https://github.com/mozilla/glean-dictionary/blob/main/etl/glean_etl.py#L130)) and it is updated daily, it should 99% track the latest release version and change when there's a new release.

> When expires == latest_fx_release_version doesn't that mean the item is expired?

Ohh I forgot that part. I'll change this to `Number(item.latest_fx_release_version) >= item.expires;` instead. Thanks for catching it!

edugfilho: np thanks for the explanation!

edugfilho (approved):

PR #2227, opened 2022-11-16
Remove y-axis formatters of memory distributions

### LINKED ISSUE #2224: Glean memory distribution scales are off when the unit is not `bytes`
When graphing Glean memory distributions, the scale seems to be off:
   - [This graph](https://glam.telemetry.mozilla.org/fenix/probe/places_manager_db_size_after_maintenance/explore?aggType=avg&currentPage=1) should be on the order of megabytes, but GLAM displays `G`
   - [This graph](https://glam.telemetry.mozilla.org/fenix/probe/glean_upload_pending_pings_directory_size/explore?aggType=avg&currentPage=1) should be on the scale of kilobytes, but GLAM display `M`

I believe this only happens when the unit is not bytes and that GLAM assumes the raw values are in the unit (kilobytes in the above cases), but Glean always stores the raw values as bytes.

Iinh:
Since Glean already stores data for memory distributions in bytes, we don't need to do the conversion on the front-end for different units.

Fixes #2224

CleanShot 2022-11-16 at 16 06 14@2x


edugfilho (approved):
lgtm; hooray for less code -> less bugs

PR #2232, opened 2022-11-22
Remove GLAM SQL template

Iinh:
Remove GLAM SQL tab since it's not really useful.


src/routing/pages/probe/GleanDetails.svelte
```diff
@@ -285,7 +284,6 @@
   </div>
   <div class="probe-details-download">
     <div class="drawer-section drawer-section--end">
-      <SqlModal />
```
Iinh, line 288: Remove "View SQL query" feature for Glean probes for now, as without the Glam SQL there's nothing to show here. Once the SQL queries are done we'll add it back up.

edugfilho (approved):
lgtm!

PR #2239, opened 2022-11-30
Improve custom distribution check

Iinh:
Explicitly state which metrics we support by having a separate list of the official metric names. 

Currently we rely on another list (probeView, used to determine which graph to display) which only checks for custom_distribution_linear or custom_distribution_exponential, thus excludes all custom_distribution -- the metric type we get back from the glean dictionary.


edugfilho (approved):
looks good. Thank you for working on that!

PR #2258, opened 2022-12-20
Bug 1804769: import skips rows with nul characters

edugfilho:
Fixes 1804769

This is a quickfix to skip rows containing nul characters coming from one of [ipc.sent_messages.parent_inactive](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/ipc_sent_messages_parent_inactive) `key`s . I decided to skip the rows instead of trying to fix them because there were only 2 samples submitted with such an error. 

I tried to import the same file locally and got the same error described in the bug. 

After fixing, here's the output of my local tests: 
Screen Shot 2022-12-19 at 7 28 45 PM

*Instead of trying to sanitize every file, notice the fix only runs in case there's a problem with the file.

### DIFF
--- a/glam/api/management/commands/import_glean_aggs.py
+++ b/glam/api/management/commands/import_glean_aggs.py
@@ -1,13 +1,16 @@
 import datetime
+import io
 import os
 import re
 import tempfile
+
 
 from django.apps import apps
 from django.core.management.base import BaseCommand
 from django.db import connection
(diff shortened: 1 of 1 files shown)


edugfilho:
After this lands we need to deploy the new GLAM image with the fix and update the DAG job to use it.

akkomar (approved):

PR #2379, opened 2023-06-12
Add non-normalized aggregations to backend

edugfilho:
Notice the migration script was done in a way that has no rollback capability. This was necessary to enable view rotation instead of deleting the current view and recreating a new one at the end, which would generate a larger downtime.

I will commit a reversible version of this migration script once this one runs, just so in the future whoever needs it can rollback to a migration prior to this one. Such rollback will generate downtime.


glam/api/migrations/0025_non_normalized_aggregations.py
```diff
@@ -0,0 +1,81 @@
...
+sql_rename = []
+sql_drop = []
+for channel in constants.CHANNEL_NAMES.values():
+    sql_create.extend(
+        [
+            f"CREATE MATERIALIZED VIEW view_glam_desktop_{channel}_aggregation_2 AS SELECT * FROM glam_desktop_{channel}_aggregation;",
+            f"CREATE UNIQUE INDEX ON view_glam_desktop_{channel}_aggregation_2 (id);",
+            f"CREATE INDEX ON view_glam_desktop_{channel}_aggregation_2 (version);",
+            f"CREATE INDEX ON view_glam_desktop_{channel}_aggregation_2 USING HASH (metric);",
+            f"CREATE INDEX ON view_glam_desktop_{channel}_aggregation_2 (os);",
```
Iinh: can we do `v2` instead of just 2, so it matches the naming pattern in ETL too.

Iinh:
> Notice the migration script was done in a way that has no rollback capability. This was necessary to enable view rotation instead of deleting the current view and recreating a new one at the end, which would generate a larger downtime.
> 
> I will commit a reversible version of this migration script once this one runs, just so in the future whoever needs it can rollback to a migration prior to this one. Such rollback will generate downtime.

To be honest I'm not sure if I understand the implications of this, but I don't want to block this effort. IIUC the plan is to test the new changes with our dev database first so at least this can easily be reverted should anything go awry. Also, I figured we should loop in our SRE contact @mikaeld so he's aware of upcoming changes to GLAM prod database.

Iinh (approved):

mikaeld (commented):
From my discussion with @edugfilho this morning, my understanding is that we don't have a high degree of confidence this can be pulled successfully within a single migration. I strongly suggest against doing a one-way migration and altering the migration scripts after the fact to add a path backward. This will only lead to confusion if we need to roll-back at some point.

I would propose breaking down this migration in 3 different migrations to have more control over what can potentially go wrong and avoid being left in a state where glam is down and we have tedious manual DB operations to get it back up. 

We could split into the following distinct migration scripts (each with a roll-back path)
1. Add fields and create new materialized view with `v2` suffix
2. Rename original view and remove `v2` suffix from new view
3. Drop the old view

This is roughly identical to what is being done here, but we'd get the added benefit of being able to validate in between steps. Rolling back the old view (migration 3) would likely take a while and incur downtime, but that's unavoidable as far as I can see. We'd also need to coordinate Glam ETL DAG changes with migration 2.

Any thoughts on this proposal?

edugfilho:
> From my discussion with @edugfilho this morning, my understanding is that we don't have a high degree of confidence this can be pulled successfully within a single migration. I strongly suggest against doing a one-way migration and altering the migration scripts after the fact to add a path backward. This will only lead to confusion if we need to roll-back at some point.
> 
> I would propose breaking down this migration in 3 different migrations to have more control over what can potentially go wrong and avoid being left in a state where glam is down and we have tedious manual DB operations to get it back up.
> 
> We could split into the following distinct migration scripts (each with a roll-back path)
> 
>     1. Add fields and create new materialized view with `v2` suffix
> 
>     2. Rename original view and remove `v2` suffix from new view
> 
>     3. Drop the old view
> 
> 
> This is roughly identical to what is being done here, but we'd get the added benefit of being able to validate in between steps. Rolling back the old view (migration 3) would likely take a while and incur downtime, but that's unavoidable as far as I can see. We'd also need to coordinate Glam ETL DAG changes with migration 2.
> 
> Any thoughts on this proposal?

@mikaeld for each one of the migration scripts, can you also describe what's their corresponding `reverse` operations?

mikaeld:
~Also, as a general advice, consider doing migrations within the scope of a transaction to ensure atomic changes. Right now, if any step fails during the migration process, it could leave the db in an inconsistent state.~

~Also, `create` and `rename` happen in two different transactions, it could lead to data loss for any operation happening in between those two statements.~ Reading Django's doc, [it seems transactions are automatically enabled for PG](https://docs.djangoproject.com/en/4.2/topics/migrations/#transactions)

Did you consider creating a temporary table from the materialized view as a mean to store data during the transaction? If performance on the materialized view is an issue, the temporary table could help with that.

Code to be tested :point_down: 
```sql
CREATE TEMPORARY TABLE temp_materialized_view_{channel}_aggregation AS 
SELECT * FROM glam_desktop_{channel}_aggregation;

DROP MATERIALIZED VIEW glam_desktop_{channel}_aggregation;

CREATE MATERIALIZED VIEW glam_desktop_{channel}_aggregation AS
SELECT *,
            NULL AS non_norm_histogram,
            NULL AS non_norm_percentiles
FROM temp_materialized_view_{channel}_aggregation;

INSERT INTO glam_desktop_{channel}_aggregation
SELECT * FROM temp_materialized_view_{channel}_aggregation;

DROP TABLE temp_materialized_view_{channel}_aggregation
```
It's been a while since I've done DBA work so bear with me, but would something like this solve your issue?

glam/api/migrations/0025_non_normalized_aggregations.py
edugfilho: the _2 here is not related to the _v2 in the ETL table name. This is just a temporary name. I could use `_temp` for it  but it could be interpreted as a temporary Materialized View instead of a temporary name

edugfilho:
@mikaeld Your SQL gets close. I was reflecting on it when I came across [RunSQL.noop ](https://til.simonwillison.net/django/migrations-runsql-noop) and realized I could use it. So I update this PR with a reversible version of the migration script.

I'm still to test it locally but the logic is sound to me. Can you take a look, please? It avoids downtime both ways and no need to further update it. Let me know what you think

edugfilho:
I made changes to my migration code. Tested it back and forth many times for migration reversibility:

* Imported data (without non-normalized aggs) into the db before migration
*  Ran migration
* Imported data (with non-normalized aggs)
* Reverted migration, then migrated forward again.. 

and no import or DB errors.

PR #2488, opened 2023-09-11
Distribution comparison

edugfilho:
Adds Distribution comparison to the context menu
Screenshot 2023-09-12 at 12 52 33 PM

Screenshot 2023-09-12 at 12 52 59 PM


src/components/explore/DistributionChart.svelte
```diff
@@ -0,0 +1,66 @@
...
+  const barWidth = bucketWidth - spaceBetweenBars
+</script>
+<style>
+  .hovd {
+    opacity: 0;
+  }
+  .hovd:hover {
+    opacity: 0.3;
+    z-index: 999;
+  }
```
Iinh, line 44: Let's add some padding between the y-axis labels, they look too close together.

CleanShot 2023-09-14 at 13 50 45@2x

Iinh: nit: could we make the modal only take up 90 view width? right now it's taking up full width and i think it's less UX friendly

src/components/explore/DistributionComparisonGraph.svelte
```diff
@@ -0,0 +1,59 @@
...
+      style: 'percent',
+      maximumFractionDigits: 2,
+    }).format(t);
+
+  let bins = density.map((d) => d['bin']);
+
+  let allTicks = Array.from(Array(topTick * 100 + tickIncrement).keys())
+    .filter((v) => v % tickIncrement == 0)
+    .map((v) => v / 100);
+  console.log(allTicks)
```
Iinh: let's remove this console.log

Iinh (changes requested):
The modal and bar chart look really good, thanks for working on this. I left some comments regarding the UI. 

Running `npm run format` should fix the failing test as it seems like a lint problem.

src/components/explore/DistributionChart.svelte
edugfilho: The labels are drawn by graph-paper and it looks like they don't have enough space to be displayed side-by-side because of the amount of buckets of that histogram. One solution is to skip some of them, but I don't think graph-paper supports that

Iinh (approved):
2 things:
- this feature currently only works with exponential histograms. other types such as count, enumerated, Glean metrics, the feature is not available. 
- with histograms with many data points such as `memory_unique`, the y-axis ticks are still very noisy

r+ for now because I do think it's better to land this PR sooner than later so we can test the changes with the dev database.

edugfilho:
> 2 things:
> 
>     * this feature currently only works with exponential histograms. other types such as count, enumerated, Glean metrics, the feature is not available.
> 
>     * with histograms with many data points such as `memory_unique`, the y-axis ticks are still very noisy
> 
> 
> r+ for now because I do think it's better to land this PR sooner than later so we can test the changes with the dev database.

You're right, and thank you for reviewing.

A few points that are in line with your comments:
- The feature is only displayed: 
  - when the violin plot is displayed
  - for firefox legacy probes
- I tried to make the bottom ticks less noisy but I didn't find a clear path to do so with the current way we're leveraging graph-paper.
- I'm glad we share the same perspective regarding having this feature out sooner

PR #2510, opened 2023-10-05
Make sure non-normalized data is passed correctly

Iinh:
Fixing an error detected during pairing yesterday where non-normalized data was not passed down correctly after changing the dropdown option.


src/components/explore/AggregationsOverTimeGraph.svelte
```diff
@@ -202,13 +205,17 @@
...
     ) {
       visiblePercentiles.forEach((p) => {
-        yData = yData.concat([...data.map((arr) => arr[percentileName][p])]);
+        yData = yData.concat([
+          ...data.map((arr) => arr[percentileName] && arr[percentileName][p]),
+        ]);
       });
       // use transformed data for Android metrics
-      if (data[0].transformedPercentiles) {
+      if (data[data.length - 1].transformedPercentiles) {
```
Iinh, line 214: we use lengh - 1, instead of 0, because non-normalized data is added later. so the first entry of data array usually doesn't have non-normalized data available

edugfilho (approved):
lgtm. Thank you!

PR #2511, opened 2023-10-05
Fix sample chart not showing due to bad data

Iinh:
Sample chart was not working on dev and prod due to bad sample data (`null`) that got in between April 9 and May 8 this year.


edugfilho (approved):
Thank you for fixing this!

PR #2513, opened 2023-10-05
Adapt non-norm histograms to distribution comparison chart

edugfilho:
This fixes a bug preventing the Distribution Comparison chart from displaying non-normalized histograms


Iinh (approved):
r+ pending adding changes that will speed up page rendering, as referenced in #2576.

edugfilho:
> r+ pending adding changes that will speed up page rendering, as referenced in #2576.

I made a comment in that ticket, upon revisiting the outcome of our efforts yesterday I don't think we should merge that

PR #2530, opened 2023-10-18
Fix scalar percentage values in table view

Iinh:
Fix this error:

CleanShot 2023-10-25 at 10 23 59@2x

### DIFF
--- a/src/components/table/TableView.svelte
+++ b/src/components/table/TableView.svelte
@@ -17,6 +17,7 @@
     timecode,
     formatPercent,
   } from '../../utils/formatters';
+  import { convertValueToPercentage } from '../../utils/probe-utils';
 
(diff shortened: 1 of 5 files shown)


src/components/table/TableView.svelte
```diff
@@ -32,6 +33,9 @@
   export let pageSize = 10;
   export let bucketTypeLabel = 'Categories';
   export let bucketOptions;
+  export let densityMetricType;
+
+  console.log('this is densitymetric', densityMetricType);
```
edugfilho:
```suggestion
```

edugfilho (changes requested):

edugfilho (approved):
thanks for fixing this!

PR #2568, opened 2023-10-31
Distribution comparison probe info and chart resize

edugfilho:
Adds probe info and adapts distribution comparison chart on window resize
https://github.com/mozilla/glam/assets/5804462/1fcd101c-611b-413b-bf40-3871c419739d


Iinh (approved):
This looks so good 👏 Thank you!

PR #2576, opened 2023-11-02
Speed up CompareDistribution component rendering

Iinh:
If this works locally for you, then we can update #2513 with it.


edugfilho (commented):
We tested this yesterday and in its actual state it does not update the Distribution Comparison chart data, and when make the chart reactive to data updates, the same problem reappears (slowness/out of memory when opening the Distribution Comparison chart). So I think we should not merge this.

PR #2577, opened 2023-11-03
Fix issues preventing glean probes from being visualized (dev)

edugfilho:
Most issues derived from the assumption glean had non normalized data, but it doesn't.

### DIFF
--- a/src/components/explore/QuantileExplorerView.svelte
+++ b/src/components/explore/QuantileExplorerView.svelte
@@ -173,9 +173,6 @@
               comparisonKeyFormatter={(perc) => `${perc}%`}
               yScaleType={probeType === 'log' ? 'scalePoint' : 'linear'}
               {yDomain}
-              )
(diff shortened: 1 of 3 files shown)


src/components/explore/QuantileExplorerView.svelte
```diff
@@ -173,9 +173,6 @@
               comparisonKeyFormatter={(perc) => `${perc}%`}
               yScaleType={probeType === 'log' ? 'scalePoint' : 'linear'}
               {yDomain}
-              )
-              ),
-              ]}
```
Iinh, line 178: @edugfilho why was this removed?

edugfilho: Seems like this was leftover from another PR. None of those characters were closing anything. A local lint error on line 177 drew my attention to this.

Iinh (approved):

PR #2581, opened 2023-11-06
Distribution comparison chart supports Glean probes

edugfilho:
This adds support to Glean probes to the Distribution Comparison chart.
I tested this with a few glean probes and the chart does not look awesome, due to a couple things:
* Seems like different builds' histograms have different number of buckets, which prevents both charts proper alignment and leads to confusion during comparison. This effect seems to be already present in the violin plot. This seems to be a data problem.
* A high number of buckets makes the x-axis labels overlap and impossible to read. This can be addressed in the visualization components, but it's hard because AFAIK [graph-paper](https://github.com/graph-paper-org/graph-paper) does not allow the x-axis to skip labels or anything that would help accomplish this
Screenshot 2023-11-06 at 4 18 40 PM

### DIFF
--- a/src/components/ChartContextMenu.svelte
+++ b/src/components/ChartContextMenu.svelte
@@ -95,7 +95,9 @@
   const canCompareDistributions = function () {
     return (
       ['firefox', 'fog'].includes($store.product) &&
-      ['histogram', 'scalar'].concat(GLEAN_METRICS).includes($store.probe.type) &&
+      ['histogram', 'scalar']
+        .concat(GLEAN_METRICS)
+        .includes($store.probe.type) &&
       !!document.getElementById(distViewButtonId)
     );
   };
--- a/src/components/ChartContextMenu.svelte
+++ b/src/components/ChartContextMenu.svelte
@@ -5,6 +5,7 @@
   import ZoomIn from './icons/ZoomIn.svelte';
   import Graphs from './icons/Graphs.svelte';
   import BarGraph from './icons/BarGraph.svelte';
+  import { SUPPORTED_METRICS as GLEAN_METRICS } from '../config/glean';
 
   export let data;
   export let x;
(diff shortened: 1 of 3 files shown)


Iinh:
> Seems like different builds' histograms have different number of buckets, which prevents both charts proper alignment and leads to confusion during comparison. This effect seems to be already present in the violin plot. This seems to be a data problem.

Not seeing this drastic difference on my local machine, but granted I'm connected to dev db, so guessing it might be less of a problem if you have more data points?

> A high number of buckets makes the x-axis labels overlap and impossible to read. This can be addressed in the visualization components, but it's hard because AFAIK [graph-paper](https://github.com/graph-paper-org/graph-paper) does not allow the x-axis to skip labels or anything that would help accomplish this

(sorry for over-taking this PR) the latest commit addresses the above problem by checking if an array has more than 20 values, then only show a selected number of values instead of all, which should solve the overcrowded-ness problem.

CleanShot 2023-11-07 at 16 48 01@2x

Iinh (approved):

edugfilho:
> > Seems like different builds' histograms have different number of buckets, which prevents both charts proper alignment and leads to confusion during comparison. This effect seems to be already present in the violin plot. This seems to be a data problem.
> 
> Not seeing this drastic difference on my local machine, but granted I'm connected to dev db, so guessing it might be less of a problem if you have more data points?
> 
> > A high number of buckets makes the x-axis labels overlap and impossible to read. This can be addressed in the visualization components, but it's hard because AFAIK [graph-paper](https://github.com/graph-paper-org/graph-paper) does not allow the x-axis to skip labels or anything that would help accomplish this
> 
> (sorry for over-taking this PR) the latest commit addresses the above problem by checking if an array has more than 20 values, then only show a selected number of values instead of all, which should solve the overcrowded-ness problem.
> CleanShot 2023-11-07 at 16 48 01@2x

Don't be sorry for doing such an amazing job! Thank you!

PR #2619, opened 2023-12-06
Make sure non-normalized data integrate well with categorical probes

Iinh:
Categorical probe types, including:
```
    'histogram-boolean': 'categorical',
    'histogram-categorical': 'categorical',
    'histogram-count': 'categorical',
    'histogram-enumerated': 'categorical',
    'histogram-flag': 'categorical',
    'scalar-boolean': 'categorical',
```

are currently not functioning properly with non-normalized data due to three main issues:
1.	The proportion data received from ETL is in raw form and has not yet been transformed into a proportion format.
2.	There are still instances where the names need to be updated, such as ﻿proportions versus ﻿non_norm_proportions, ﻿counts versus ﻿non_norm_counts.
3.	Filtering out empty non-normalized data correctly is problematic since it is not always an empty string ﻿""; sometimes it can be an empty object.

This pull request addresses the aforementioned problems. I have included comments to explain the code.

### DIFF
--- a/src/components/explore/ProbeExplorer.svelte
+++ b/src/components/explore/ProbeExplorer.svelte
@@ -1,4 +1,5 @@
 <script>
+  import { isEmpty } from 'lodash';
   import { writable } from 'svelte/store';
 
   import { window1D } from '@graph-paper/core/utils/window-functions';
@@ -48,7 +49,7 @@
     // data which caused the graph to break.
     // so, we filter out these empty data points.
     normType === 'non_normalized'
-      ? normData.filter((d) => d.non_norm_histogram !== '')
+      ? normData.filter((d) => !isEmpty(d.non_norm_histogram))
       : normData;
 
   let data = filterData(
--- a/src/components/explore/ProportionExplorerView.svelte
+++ b/src/components/explore/ProportionExplorerView.svelte
@@ -26,7 +26,12 @@
     gatherAggregationTypes,
   } from '../../utils/probe-utils';
   import { numericStringsSort } from '../../utils/sort';
-  import { getHistogramName, numHighlightedBuckets } from '../../config/shared';
+  import {
+    getHistogramName,
+    getProportionName,
+    getCountName,
+    numHighlightedBuckets,
+  } from '../../config/shared';
 
   export let aggregationLevel = 'build_id';
   export let data;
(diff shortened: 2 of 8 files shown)


src/components/explore/ProbeExplorer.svelte
```diff
@@ -48,7 +49,7 @@
     // data which caused the graph to break.
     // so, we filter out these empty data points.
     normType === 'non_normalized'
-      ? normData.filter((d) => d.non_norm_histogram !== '')
+      ? normData.filter((d) => !isEmpty(d.non_norm_histogram))
```
Iinh, line 52: when there's no non_normalized data, ie empty `non_norm_histogram`, `d.non_norm_histogram` can be either an empty string or object depending on the probe type => using `isEmpty()` here to catch all cases.

src/components/explore/ProportionExplorerView.svelte
```diff
@@ -52,28 +57,27 @@
...
   let currentAggregation = aggregationTypes[0];
 
+  function filterResponseData(d, agg, key) {
+    return d.filter(
+      (di) => di.client_agg_type === agg && di.metric_key === key
+    );
+  }
+
+  $: selectedData = filterResponseData(data, currentAggregation, currentKey);
+
```
Iinh, line 67: moving the `selectedData` initialization up here so i can use it below ⬇️

src/utils/transform-data.js
```diff
@@ -92,7 +93,12 @@ export function checkForTotalUsers(draft) {
 export function addProportion(draft) {
   // requires point.histogram.
   draft.proportions = { ...draft.histogram };
-  draft.non_norm_proportions = { ...draft.non_norm_histogram };
+
+  // draft.non_norm_histogram is not in proportion format
+  // like draft.histogram so we need to convert it here
+  draft.non_norm_proportions = convertValueToProportions(
+    draft.non_norm_histogram
+  );
```
Iinh, line 101: I noticed that for categorical probes, we need the values in proportion format. eg.

`const proportions = { 1: 0.4, 1: 0.3, 2: 0.2, 3: 0.1};`

However the values from for non_normalized data are raw and have not been transformed yet, so we need to add it here.

edugfilho (approved):
LGTM! Thanks!

PR #2633, opened 2023-12-12
Increase search result limit

Iinh:
When prepping for the presentation demo, we noticed that keeping result limit to only 30 filters out more probes that we would have wanted.


Iinh:
pinging @edugfilho again to r? this PR 😅

edugfilho (approved):

PR #2670, opened 2024-01-19
ADR - Serve data from BQ

edugfilho:
Accepting feedback on this ADR, with code attached.

Moving this forward we probably need to coordinate with DSRE to make sure GLAM has proper BQ permissions.


scholtzan:
I'd recommend that whatever service account is used for accessing BigQuery only has access to the tables that should be exposed. There seem to be a bunch of tables in `glam_etl` that might contain data we'd not want users to get access to if somehow the account credentials are leaked or the API accidentally allows access to them somehow.

edugfilho:
> I'd recommend that whatever service account is used for accessing BigQuery only has access to the tables that should be exposed. There seem to be a bunch of tables in `glam_etl` that might contain data we'd not want users to get access to if somehow the account credentials are leaked or the API accidentally allows access to them somehow.

I agree but I'd like to strike some middle ground there in the likely case we'll need to add more tables to the `glam_etl` dataset for a new product, for example. In that case I wouldn't like to get DSRE involved, so if BQ supports asterisk for access control that may be a good idea. Or even creating a whole new dataset only for prod tables and allowing GLAM to that dataset only.

glam/api/views.py
```diff
@@ -397,10 +640,13 @@ def aggregations(request):
             )
         )
 
+    data_source = "BigQuery"
```
edugfilho: Rollback flag

edugfilho:
I initially intended to create a feature flag for this and select a number of users to rollout this to, but it proved too complicated since a full rollback is as simple as switching a flag and re-deploying.
So I intend to deploy this in its current state, which means the moment it hits dev all requests in that environment will be fetching data from the BigQuery prod tables, and we can test it out. If things look good then we do stage, then prod.

Rollback is as simple as switching the `data_source` string in `glam/api/views.py` back to `Postgres` and re-deploying.

So my plan is to deploy this to dev on Jan 29.

@mikaeld requesting your input for the above deployment/rollback plan, because if we want to rollback prod it will involve a prod re-deployment. But just button pressing as usual. 

Also, we want GLAM to have read permissions to the following tables and eventually other tables (e.g. in case of new products getting added), while taking into account Anna's comment below. I was thinking of a dataset-level access, so it doesn't involve DSRE adding access every time for a new table. Would you have a different recommendation?

> I'd recommend that whatever service account is used for accessing BigQuery only has access to the tables that should be exposed. There seem to be a bunch of tables in `glam_etl` that might contain data we'd not want users to get access to if somehow the account credentials are leaked or the API accidentally allows access to them somehow.

Here are the tables to which GLAM currently needs access:
- moz-fx-data-shared-prod.telemetry.buildhub2
- moz-fx-data-shared-prod.telemetry_derived.latest_versions_v2

Desktop (legacy):
- moz-fx-data-glam-prod-fca7.glam_etl.glam_desktop_release_aggregates_v1
- moz-fx-data-glam-prod-fca7.glam_etl.glam_desktop_nightly_aggregates_v1
- moz-fx-data-glam-prod-fca7.glam_etl.glam_desktop_beta_aggregates_v1
Desktop(Glean)
- moz-fx-data-glam-prod-fca7.glam_etl.glam_fog_release_aggregates_v1
- moz-fx-data-glam-prod-fca7.glam_etl.glam_fog_nightly_aggregates_v1
- moz-fx-data-glam-prod-fca7.glam_etl.glam_fog_beta_aggregates_v1

Android:
- moz-fx-data-glam-prod-fca7.glam_etl.glam_fenix_release_aggregates_v1
- moz-fx-data-glam-prod-fca7.glam_etl.glam_fenix_nightly_aggregates_v1
- moz-fx-data-glam-prod-fca7.glam_etl.glam_fenix_beta_aggregates_v1

glam/api/views.py
```diff
@@ -46,7 +61,14 @@ def updates(request):
...
 
-def get_firefox_aggregations(request, **kwargs):
+def get_firefox_aggregations(source, request, **kwargs):
+    if source == "BigQuery":
+        return get_firefox_aggregations_from_bq(request, **kwargs)
+    else:
+        return get_firefox_aggregations_from_pg(request, **kwargs)
+
+
+def get_firefox_aggregations_from_pg(request, **kwargs):
```
mikaeld, line 61: I'm not sure if Django supports BigQuery with [database routers](https://docs.djangoproject.com/en/5.0/topics/db/multi-db/), but in case it doesn't, the client/session object should leverage dependency injection rather than creating an helper function to route a call to partially duplicated functions.

glam/api/views.py
```diff
@@ -27,6 +30,18 @@
...
+
+    _instance = None
+
+    def __new__(cls):
+        if not cls._instance:
+            cls._instance = super(GlamBigQueryClient, cls).__new__(cls)
+            cls._instance.client = bigquery.Client(project="moz-fx-data-glam-prod-fca7")
+        return cls._instance.client
+
+
```
mikaeld: Considering this [gunicorn config](https://github.com/mozilla/glam/blob/e42096fdbc1ad1be2644be462c7aa875ef99da6c/Dockerfile#L82-L84) using `Client` as a context manager seems like a better alternative to ensure sane instance lifecycle.

mikaeld:
Table-level access is preferable to dataset-level. You can use the existing `workgroup:dataops-managed/glam` in bqetl metadata for all the tables you listed.

glam/api/views.py
edugfilho: Database routers aren't supported. And the `_from_pg` function is there only to ensure easy rollback. Therefore this helper function will go away together with the `_from_pg` variant once we are confident we're sticking to BQ. 

The goal is to be able to do a quick switch and investigate any problems later locally in case of some disruption in prod caused by this change.

glam/api/views.py
edugfilho: I agree, despite not exactly being able to thing of a case where this could get messy. In any case, that's been done. Thanks!

mikaeld: I mean `Client` already [implements a context manager.](https://github.com/googleapis/python-bigquery/blob/62490325f64e5d66303d9218992e28ac5f21cb3f/google/cloud/bigquery/client.py#L4129-L4133), why not use it directly? More specifically, your implementation does not close the Client connection.

mikaeld: I'm also curious why use a singleton at all? Did you notice significant delay when instantiating a connection?

edugfilho: Regarding the singleton, I didn't notice a significant delay and I come to realize it was a mistake. I also didn't know BQ client implemented context manager, so TYVM for this. How does that look now?

glam/api/views.py
```diff
@@ -281,6 +413,105 @@ def get_glean_aggregations(request, **kwargs):
...
+    """
+    job_config = bigquery.QueryJobConfig(
+        query_parameters=[
+            bigquery.ScalarQueryParameter("metric", "STRING", probe),
+            bigquery.ScalarQueryParameter("ping_type", "STRING", ping_type),
+            bigquery.ScalarQueryParameter("os", "STRING", os),
+        ]
+    )
+    with bigquery.Client(project=GLAM_BQ_PROD_PROJECT) as client:
+        query_job = client.query(query, job_config=job_config)
```
mikaeld, line 479: Usually, you would use dependency injection to pass a fully instantiated client object rather than create it in the function. A couple of advantages: using the same client object for multiple queries within a route, loose coupling means it's way easier to unit test because you can inject a mocked client, etc.

mikaeld: In your case, since PG leverages Django directly and does not require DI, then you could create your BQ Client [here](https://github.com/mozilla/glam/blob/d3919d22afe0414a3f4d2d403715f7cf8ec15e46/glam/api/views.py#L311) and [here](https://github.com/mozilla/glam/blob/d3919d22afe0414a3f4d2d403715f7cf8ec15e46/glam/api/views.py#L56), and pass it as a function argument. That way, your function becomes easier to test.

mikaeld (approved):

glam/api/views.py
edugfilho: 100%
I'll land the currently code and work on this soon

PR #2684, opened 2024-02-01
Use appropriate BQ project according to env. Random probes fetch data from BQ

edugfilho:


Iinh (approved):
LGTM. r+ pending passing tests if possible.

PR #2687, opened 2024-02-05
Remove state reset, we only need to reset ref point

### LINKED ISSUE #2683: [stage] Can't visualize Android probes after selecting from the search bar
Steps to reproduce:
- Go to https://stage.glam.nonprod.dataops.mozgcp.net
- On the drop-down on the right of the search bar, select Firefox Android
- Search for any probe and select it
- Notice that the url still has `/firefox/` instead of `/fenix/` the page keeps loading forever

Works fine in prod.

Iinh:
Fixes #2683

### DIFF
--- a/src/components/search/SearchResults.svelte
+++ b/src/components/search/SearchResults.svelte
@@ -122,9 +122,9 @@
   };
 
   const onClick = () => {
-    // we need to clear the query state here, otherwise the query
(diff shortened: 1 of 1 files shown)


edugfilho (approved):
LGTM. Thanks for fixing it!

PR #2738, opened 2024-03-06
Add Matrix links and update Slack links where needed

Iinh:
As we're taking GLAM public, we should remove Slack links where not needed, and replace it with a link to our Matrix instance.

Front page changes:

CleanShot 2024-03-06 at 10 58 57@2x


src/routing/wrappers/Layout.svelte
```diff
@@ -55,9 +54,10 @@
       <a
         target="_blank"
         class="temporary-link-button"
-        href="https://mozilla.slack.com/archives/CB1EQ437S"
+        href="https://matrix.to/#/#datatools:mozilla.org"
```
edugfilho, line 57: I think we want to have both logos with links under feedback (not necessarily in a dropdown), no?

Iinh: If I remember correctly we decided for this upcoming public release we can keep this for now, and to turn this into a drop down in the next iteration (which we'll need to open a ticket for this.) The hero text that mentions both Slack and Matrix will suffice for now. I think in the last triage we'd agreed that having both links there will make the UI too noisy for that corner. Let me know what you think.

edugfilho: Sounds good to me!

edugfilho (approved):

PR #2742, opened 2024-03-07
Landing page shows cached, most used probes instead of random

edugfilho:
Makes the landing page more useful and faster \o/


glam/api/migrations/0028_cache_table.py
```diff
@@ -0,0 +1,19 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+            ('api', '0027_fix_fog_user_count_after_sampling'),
+        ]
+
+    operations = [
+        migrations.RunPython(
+            code=create_cache, reverse_code=migrations.RunPython.noop
```
edugfilho: We need to create a cache table on PG in order to cache the most used probes

edugfilho: Also, I don't want to destroy the cache table as a reverse operation, it's better to have it be intentional.

src/state/api.js
```diff
@@ -14,11 +14,7 @@ const DEFAULT_SEARCH_RESULTS_LIMIT = 100; // maximum number of results to show
 
 export async function getRandomProbes(numProbes, process) {
   const data = await fetch(randomProbeURL, {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
-    },
-    body: JSON.stringify({ n: numProbes, process }),
+    method: 'GET',
```
edugfilho: GET is the right method for this. Besides, django caching only works with HEAD or GET.

glam/api/views.py
```diff
@@ -690,21 +691,88 @@ def _get_random_probes(data_source, random_percentage, limit):
...
+            AND version = selected_version.v
+        GROUP BY metric
+        LIMIT {limit}"""
+
+    with bigquery.Client() as client:
+        aggs = client.query(query)
+    return aggs
+
+
+@api_view(["GET"])
```
edugfilho, line 766: GET is the right method for this. Besides, django caching only works with HEAD or GET.

PR #2816, opened 2024-05-08
Match key state handling to onClick

Iinh:
Update key behaviors to match onclick. Thanks @rafrombrc for finding this bug!


edugfilho (approved):
LGTM! thanks!

PR #2779, opened 2024-04-02
#2693 Improves Distribution Comparison for categorical probes

edugfilho:
Fix #2693 
Labels are properly displayed on chart
Only selected categories are displayed on chart

Fixed issue with Distribution Comparison showing wrong values for Scalar probes.

https://github.com/mozilla/glam/assets/5804462/21982b70-9351-471f-a10a-ac59ccd2dd3f

PR #2817, opened 2024-05-14
Fix glean metrics not showing after fix for Categorical probes

edugfilho:
This fixes the bug introduced in https://github.com/mozilla/glam/pull/2779, which searches for a field that's not present into Glean metrics' metadata.

### DIFF
--- a/src/components/DistributionComparisonModal.svelte
+++ b/src/components/DistributionComparisonModal.svelte
@@ -14,14 +14,14 @@
 
   let normalized = $store.productDimensions.normalizationType === 'normalized';
   let probeType = $store.probe.type;
-  let isCategoricalProbe = $store.probe.details && $store.probe.details.kind === 'categorical';
+  let isCategoricalProbe =
+    $store.probe.details && $store.probe.details.kind === 'categorical';
   let cumulative = false;
-  let activeCategoricalProbeLabels =
-    isCategoricalProbe
-      ? $store.probe.details.labels.filter((l) =>
-          $store.activeBuckets.includes(l)
-        )
-      : [];
+  let activeCategoricalProbeLabels = isCategoricalProbe
+    ? $store.probe.details.labels.filter((l) =>
+        $store.activeBuckets.includes(l)
+      )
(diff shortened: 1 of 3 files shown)


src/components/DistributionComparisonModal.svelte
```diff
@@ -14,14 +14,14 @@
 
   let normalized = $store.productDimensions.normalizationType === 'normalized';
   let probeType = $store.probe.type;
-  let probeKind = $store.probe.details.kind;
+  let isCategoricalProbe =
+    $store.probe.details && $store.probe.details.kind === 'categorical';
```
Iinh, line 18: would this only work with Glean metrics? how about categorical histograms?

edugfilho: My understanding is we want it to be true iff it's a Legacy probe and it's categorical, no?

edugfilho: Legacy probes have `details` whereas Glean metrics don't. Categorical histograms have to have `isCategoricalProbe === true`

Iinh: ah yea thanks for the clarification. could you remind me why we don't need to check for categorical glean probes here -  distribution comparison is not available for glean probes yet?

edugfilho: I didn't think Glean had categorical probes. I did a quick look up at Glean dictionary and didn't find any. Oh, are they under a different name?

Iinh: I think Glean boolean probes have categorical view in Glam, but will need to check if we even display these. 

https://github.com/mozilla/glam/blob/f0228a3947e2f8d9bdc1dd6e5bbdd0c6e7adfc8f/src/config/glean.js#L75

edugfilho: hm, if that's the case GLAM doesn't support `boolean` probes

Iinh (approved):
Thanks for fixing this!

PR #2818, opened 2024-05-14
Search for Darwin when Mac is selected

edugfilho:
Found this while testing other stuff. GLAM erroneously reporting no data for Mac metric all along because in the data it's called Darwin.

### DIFF
--- a/src/config/glean.js
+++ b/src/config/glean.js
@@ -42,7 +42,7 @@ export const FIREFOX_ON_GLEAN = {
       values: [
         { key: '*', label: 'All OSes' },
         { key: 'Windows', label: 'Windows' },
-        { key: 'Mac', label: 'Mac' },
+        { key: 'Darwin', label: 'Mac' },
         { key: 'Linux', label: 'Linux' },
       ],
       defaultValue: '*',


src/config/glean.js
```diff
@@ -42,7 +42,7 @@ export const FIREFOX_ON_GLEAN = {
       values: [
         { key: '*', label: 'All OSes' },
         { key: 'Windows', label: 'Windows' },
-        { key: 'Mac', label: 'Mac' },
+        { key: 'Darwin', label: 'Mac' },
         { key: 'Linux', label: 'Linux' },
```
Iinh, line 46: Looking at this [ingestion code](https://github.com/mozilla/gcp-ingestion/blob/a6928fb089f1652856147c4605df715f327edfcd/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java#L52-L74), I think we're missing more than just Mac?

edugfilho: Interesting. I did find a very small amount of BSD in the data. The thing is if we want to accommodate more than one name into a single alias, we have to do that at the ETL level, otherwise the aggregations won't make sense. I say we land this so ppl can look at Mac data and open an issue for your find?

Iinh: I agree. Maybe good to use this [normalized os UDF](https://mozilla.github.io/bigquery-etl/mozfun/norm/#os-udf) at the ETL level - `mozfun.norm.os(normalized_os)`.

Iinh (approved):

src/config/glean.js
edugfilho: Perfect! Thanks for finding this. Would you mind opening the ticket with this precious info?

Iinh: done!

https://github.com/mozilla/glam/issues/2819

PR #2820, opened 2024-05-14
Validate request parameters

edugfilho:
This PR validates parameters before passing them to the sql query.

### DIFF
--- a/glam/api/views.py
+++ b/glam/api/views.py
@@ -57,10 +57,85 @@ def updates(request):
     )
 
 
+def validate_request_legacy(**kwargs):
+    # TODO: When glam starts sending "product", make it required.
+    REQUIRED_QUERY_PARAMETERS = ["channel", "probe", "aggregationLevel"]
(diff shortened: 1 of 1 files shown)


glam/api/views.py
```diff
@@ -57,10 +57,100 @@ def updates(request):
...
+        raise ValidationError(
+            "Invalid versions: {}".format(validated_data["num_versions"])
+        )
+
+    return validated_data
+
+
+def validate_request_glean(**kwargs):
+    REQUIRED_QUERY_PARAMETERS = [
+        "aggregationLevel",
```
Iinh, line 98: I didn't realize aggregationLevel in the request is in a different casing (camelCase vs snake_case) than the rest of the params.

Iinh (approved):
LGTM - Thanks for adding more data validation to our code! Out of curiosity, for context, was there any example in the past that made you think validation is necessary beyond a cautionary measure?

Issue #2985, opened 2024-10-16
Interpolate percentiles when appropriate


hotsphink:
This also came up in #1639 and #2054. The latter discussed two different solutions to two mostly different problems, and so the issue ended up being about visual smoothing over time. I didn't reuse #1639 because I didn't totally follow the discussion there (particularly @chutten's [second point](https://github.com/mozilla/glam/issues/1639#issuecomment-941378850)), and that issue might have different solutions.

We have many probes that produce bucketed samples of continuous values (eg pretty much anything that is measuring time). The percentile appears to be using a nearest-rank method. As can be seen in @jonco3's example in #2054, this produces choppy results and does not correspond to the underlying continuous distribution. (As the [wikipedia article](https://en.m.wikipedia.org/wiki/Percentile) puts it, "The simplest are nearest-rank methods... although compared to interpolation methods, results can be a bit crude.") This is especially problematic for high percentiles (95th and 99th) that are more likely to flip flop between buckets. Interpolation would make the display more sensitive to the changes we care about.

Given that we know the underlying distribution is continuous, we'd like to be able to see an interpolation between closest ranks. We're fine if it must be selected manually, or if it requires labeling the probe as continuous or whatever.

edugfilho:
I've been working on a PoC for interpolation. My first try was using a nearest-rank method and that didn't change much (the percentile lines were still choppy), then I experimented with a moving average in which the window size is 1% of the dataset on the percentiles, and that did yield some results, as expected (see video below). I invite you to check it out on [GLAM dev](https://dev.glam.nonprod.dataops.mozgcp.net/firefox/probe/gc_mark_ms/explore?), pick a different probe and tell me what you think. In the meantime I'll implement between closest ranks and put together a side-by-side comparison.

hotsphink:
Oh oops, I misread your comment as saying you had already done an interpolation between closest ranks, which was confusing because that's not what the result looked like to me. But I now see you're talking about the time-based smoothing.

I would prefer interpolation between closest buckets and *not* having any smoothing over time. Smoothing over time (the moving average) makes for nicer graphs and is a good option to have available, but it adds delay and attenuation. Given that most of the choppiness is artificial for samples of continuous values, I'd rather see a pixel based only on the buckets from the time range corresponding to that pixel, if that makes sense? If there were to be a code change that legitimately changed a value, that should show up *and* be right at the time when the change occurred. There will still be lots of wiggles and variance that will obscure any changes, but at least that's innate.

[...]

PR #2995, opened 2024-10-22
feature: Interpolate between closest ranks

edugfilho:
implements https://github.com/mozilla/glam/issues/2985

### DIFF
--- a/src/components/explore/ProbeExplorer.svelte
+++ b/src/components/explore/ProbeExplorer.svelte
@@ -60,36 +60,39 @@
     // Compute cumulative frequencies
     let cumFreq = [];
     let totalFreq = 0;
-    for (let i = 0; i < histogram.length; i++) {
+
+    for (let i = 0; i < histogram.length; i += 1) {
       totalFreq += histogram[i].value;
       cumFreq.push({ bin: histogram[i].bin, cumFreq: totalFreq });
     }
     // Normalize so we can find percentiles appropriately
-    if (totalFreq != 1) {
-      cumFreq = cumFreq.map((item) => {
-        return { bin: item.bin, cumFreq: item.cumFreq / totalFreq };
-      });
+    if (totalFreq !== 1) {
+      cumFreq = cumFreq.map((item) => ({
+        bin: item.bin,
+        cumFreq: item.cumFreq / totalFreq,
+      }));
     }
 
     // Find the interval where each percentile falls and interpolate
     const percentileValues = {};
-    for (const percentile of percentiles) {
+    for (let i = 0; i < percentiles.length; i += 1) {
+      const percentile = percentiles[i];
       let targetFreq = percentile / 100;
       if (targetFreq <= cumFreq[0].cumFreq) {
         percentileValues[percentile] = cumFreq[0].bin;
       }
-      if (targetFreq >= cumFreq.at(-1).cumFreq) {
+      if (targetFreq >= totalFreq) {
         percentileValues[percentile] = cumFreq[cumFreq.length - 1].bin;
       }
-      for (let i = 0; i < cumFreq.length - 1; i++) {
+      for (let j = 0; j < cumFreq.length - 1; j += 1) {
         if (
-          cumFreq[i].cumFreq <= targetFreq &&
-          cumFreq[i + 1].cumFreq >= targetFreq
+          cumFreq[j].cumFreq <= targetFreq &&
+          cumFreq[j + 1].cumFreq >= targetFreq
         ) {
-          let x0 = cumFreq[i].cumFreq;
-          let x1 = cumFreq[i + 1].cumFreq;
-          let y0 = cumFreq[i].bin;
-          let y1 = cumFreq[i + 1].bin;
+          let x0 = cumFreq[j].cumFreq;
+          let x1 = cumFreq[j + 1].cumFreq;
+          let y0 = cumFreq[j].bin;
+          let y1 = cumFreq[j + 1].bin;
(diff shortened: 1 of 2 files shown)


src/components/explore/ProbeExplorer.svelte
```diff
@@ -56,19 +56,71 @@
...
+
+  function getInterpPercBtnRanksForHistogram(histogram, percentiles) {
+    // Compute cumulative frequencies
+    var cumFreq = [];
+    var totalFreq = 0;
+    for (var i = 0; i < histogram.length; i++) {
+      totalFreq += histogram[i].value;
+      cumFreq.push({ bin: histogram[i].bin, cumFreq: totalFreq });
+    }
+    const normalizer = cumFreq.at(-1).cumFreq
```
hotsphink: This is just `totalFreq`, fwiw. Except in the edge case where `histogram` is empty, but then the `.cumFreq` would throw an error. (It's probably impossible?)

src/components/explore/ProbeExplorer.svelte
```diff
@@ -56,19 +56,71 @@
...
+    if (normalizer != 1) {
+      cumFreq = cumFreq.map((item) => {return {bin: item.bin, cumFreq: item.cumFreq/normalizer}})
```
hotsphink: Not that it really matters, but rather than creating a new set of objects, this could do an in-place update:
```
  const normalizer = 1 / totalFreq;
  cumFreq.forEach(item => { item.cumFreq *= normalizer });
```

src/components/explore/ProbeExplorer.svelte
```diff
@@ -56,19 +56,71 @@
...
     }
+
+    // Find the interval where each percentile falls and interpolate
+    const percentileValues = {}
+    for (const percentile of percentiles) {
+      var targetFreq = percentile / 100;
+      if (targetFreq <= cumFreq[0].cumFreq) {
+        percentileValues[percentile] = cumFreq[0].bin;
+      }
+      if (targetFreq >= cumFreq.at(-1).cumFreq) {
```
hotsphink: `if (targetFreq >= totalFreq)`

src/components/explore/ProbeExplorer.svelte
```diff
@@ -56,19 +56,71 @@
...
+        percentileValues[percentile] = cumFreq[cumFreq.length - 1].bin;
+      }
+      for (var i = 0; i < cumFreq.length - 1; i++) {
+        if (cumFreq[i].cumFreq <= targetFreq && cumFreq[i + 1].cumFreq >= targetFreq) {
+          var x0 = cumFreq[i].cumFreq;
+          var x1 = cumFreq[i + 1].cumFreq;
+          var y0 = cumFreq[i].bin;
+          var y1 = cumFreq[i + 1].bin;
+          // Linear interpolation formula
+          var percentileValue = y0 + ((targetFreq - x0) * (y1 - y0)) / (x1 - x0);
```
hotsphink: I think this works, but I'm not sure how the bin values are given. If one bucket is for values 3-7, then will `.bin` be 3, 7, or (3+7)/2=5? With exponentially increasing bucket sizes, it seems like it would make a difference. ...after more research, I think your formula is exactly right as long as the bins are given as the lowest value in each bucket, which seems like it must be the case. (Then the range of values that fall in some bucket i are [cumFreq[i].bin, cumFreq[i+1].bin) which is exactly what you're using here, so I should shut up already.)

It's a little sketchy, because the linear interpolation assumes that the values are uniformly distributed _within_ the bucket. Apparently sometimes people will use a logarithmic interpolation instead, which is for when the values are exponentially distributed within the bucket. I'm not at all sure about this stuff, but I don't think we can assume any particular distribution. And fortunately, the linear interpolation won't be very far off most other distributions. I think?

edugfilho: Your assumptions are correct. I added a note about this in the tooltip that will show next to the check-box that toggles the feature.
I also think that if we need to add different types of interpolation based on something (e.g. probe metadata) we can do that quickly now that we already have the core feature implemented

src/components/explore/ProbeExplorer.svelte
edugfilho: thank you. I appreciate the review!

src/components/explore/QuantileExplorerView.svelte
```diff
@@ -194,20 +193,20 @@
...
+                    bind:checked={interpolate}
                   />
                   <h3 for="toggleSmooth" class="data-graphic__element-title">
-                    Interpolate
+                    Interpolated
                   </h3>
                   <span
                     use:tooltipAction={{
-                      text: 'Applies a moving average to smooth out short-term fluctuations on percentile values.',
+                      text: 'Generates percentiles using the Between Closest Ranks Linear Interpolation. This can show an innacurate representation of the data if the underlying distribution is not continuous and/or the data between bins is not uniformly distributed.',
```
hotsphink, line 209: s/innacurate/inaccurate/

Also, this isn't quite right. This does not depend on the distribution of data *between* bins, only on the distribution of data *within* bins. It will actually handle any distribution between bins, which is fortunate since it's usually going to some weird multimodal exponential-ish thing.

edugfilho: oops, my bad. I'll open another PR for this. I thought you were done.

hotsphink: > oops, my bad. I'll open another PR for this. I thought you were done.

Sorry, I thought I was too.

PR #3004, opened 2024-11-06
Add Additional Delivery to Google Artifact Repo

corbanCloud:
Adds GAR (Google Artifact Repository) as a second delivery source for GLAM images

Functionality was test in a temporary branch, which produced [this artifact](https://console.cloud.google.com/artifacts/docker/moz-fx-glam-prod/us/glam-prod/glam?project=moz-fx-glam-prod) 

The pipeline steps here are intentionally not DRY'd up between the existing GCR delivery as those steps can simply be deleted after V2 Migration is complete. Additionally GAR delivery was added for both the main glam app container and the auth0 proxy which creates some additional cruft during the transition

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,6 +147,7 @@ jobs:
     executor: gcp-gcr/default
     environment:
       GOOGLE_PROJECT_ID: moz-fx-dataops-images-global
+      GLAM_PROJECT_ID: moz-fx-glam-prod
     steps:
       - checkout
       - gcp-gcr/gcr-auth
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -116,11 +117,37 @@ jobs:
               docker tag app:build "${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}:${CIRCLE_TAG}"
               docker push ${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}:${CIRCLE_TAG}
             fi
+      - run:
+          name: Prepare environment variables for OIDC authentication
+          command: |
+            echo 'export GOOGLE_PROJECT_ID="moz-fx-experimenter-prod-6cd5"' >> "$BASH_ENV"
```
whd: This is definitely incorrect copypasta. If this worked to produce the test image https://console.cloud.google.com/artifacts/docker/moz-fx-glam-prod/us/glam-prod/glam?project=moz-fx-glam-prod that's pretty interesting information and suggests we maybe don't need to set this value.

```suggestion
            echo 'export GOOGLE_PROJECT_ID="moz-fx-glam-prod"' >> "$BASH_ENV"
```

Additionally: some of these variables are available via https://app.circleci.com/settings/organization/github/mozilla/contexts/02806e84-f505-4ad2-b8b1-5eda3c797029?return-to=%2F so you don't need to set them in the project-specific environment variables if you use that context. I guess this is in a bit of a confusing state while you're doing both GCR/GAR so as long as it works :+1:

.circleci/config.yml
```diff
@@ -148,6 +175,39 @@ jobs:
           tag: $CIRCLE_TAG
           google-project-id: GOOGLE_PROJECT_ID
 
+      # this could be done by re-tagging the existing image but for ease of migration
+      # I'm duplicating it fully with the intent of removing the above steps entirely
+      # /moz-fx-glam-prod/glam-prod/glam@sha256:705f53c5511c5803c6ea58896a93a97a30dc8defdef19c5d2243c3171a13be25
+      - run:
```
whd, line 181: Do you need to do this if you're just using the `gcp-gcr`  orb? You're not actually using the GCP CLI after you set it up.  It seems like you should be able to get away with a simpler config than this that's just `gcp-gcr/build-and-push-image` with the right environment variables set and with `use_oidc` set to `true`. But if this works then that's fine.

.circleci/config.yml
```diff
@@ -148,6 +175,39 @@ jobs:
...
+          workspace-root: dockerfiles/proxy
+          path: dockerfiles/proxy
+          docker-context: dockerfiles/proxy
+          registry-url: us-docker.pkg.dev
+          google-project-id: GLAM_PROJECT_ID
+          repository: glam-prod
+          image: auth0-proxy
+          tag: $CIRCLE_SHA1
+
+      - gcp-gcr/push-image:
```
whd, line 203: Has this been tested yet? I don't see any images in https://console.cloud.google.com/artifacts/docker/moz-fx-glam-prod/us/glam-prod?project=moz-fx-glam-prod other than the `glam` image.

whd (approved):
This looks like it won't work to me for the ancillary images but I might be reading it wrong. As long as images get pushed to all the right places :+1:

PR #3008, opened 2024-11-14
Seperate GAR upload step, rename deploy to legacy-deploy

corbanCloud:
Changing CI to make my own troubleshooting easier, and move this closer to end-state at once.

All images will get uploaded to GAR, tags will be used to differentiate stage vs prod releases from master.
Dev can be handled a few ways, will depend on chats w/ Eduardo (GLAM Dev) on how he wants it to work


corbanCloud:
renaming branch for clarity in #3009

PR #3009, opened 2024-11-15
Fix GAR Authentication

corbanCloud:
(renamed branch from #3008 for posterity)

Changing CI to make my own troubleshooting easier, and move this closer to end-state at once.

All images will get uploaded to GAR, tags will be used to differentiate stage vs prod releases from master.
Dev can be handled a few ways, will depend on chats w/ Eduardo (GLAM Dev) on how he wants it to work

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,42 @@ jobs:
             - /tmp/cache/app-frontend.tar
             - /tmp/cache/app.tar
 
+  gar-upload:
+    docker:
+      - image: cimg/base:stable
+        auth:
+          username: $DOCKER_USER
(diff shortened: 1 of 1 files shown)


corbanCloud:
Ok build + tagging now works as expected, including changes needed to support the Argo workflow I have in mind for this project 

Tagging a commit will add that tag to the built image as well, Argo will be setup to pull all "latest" images into stage, then version tagged images into prod, but this will not result in a new image being created since the tag will be added to the existing image since the container SHA256 will be the same between pipelines

See screen cap from Glam prod GAR

Screenshot 2024-11-14 at 7 05 41 PM

PR #3010, opened 2024-12-02
feat: V2 migration CI improvements

corbanCloud:
- Code clean up from v2 migration, and other housekeeping
- Simplifies adding and pushing tags on images
- GAR tag+upload is blocked if a main-line semver tag is applied to a non-main branch

Image uploading and promotion workflow is intended to be as follows per request/discussion w/ @edugfilho 
- Dev will track the latest commit on the `main` branch for dev (by way of the `:latest` tag in GAR). Images are also tagged with `branch` and `commit` info incase there is a need to have dev fetch a working branch or specific past commit.
- Stage will be configured to track all releases with release candidate and similar semver tags e.g. v1.1-rc0, v1.2.3-rc4, also allowed `alpha` and `beta` tags
- Prod will be configured to track will track all stable tagged releases e.g v1.1, v4.5.6 

In order to prevent accidental production releases, the `gar-upload` step checks if the branch is the main branch AND the tag value is a non-qualified semver. If this condition is true the step logs the unexpected behavior and exits before taking any other actions incl. uploading images.

We expect the convention to be that stage releases ought to be merged to `main` first and then promoted as a release candidate by way of an `rc` tag (e.g. 1.1.1-rc1) but are not guarding against cases where that does not happen that at this time. In case a branch needed to get deployed to stage, tagging any commit with the qualifier `alpha` or `beta` would allow that while staying within semver norms.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,13 +69,18 @@ jobs:
       - run:
           name: Deploy to Google Artifact Registry
           command: | #shell
+            
+            # Exit w/ error if non-main branch is tagged with format 0.0.0
+            # per discssion w/ dev team we expect non-main branches to have alpha, beta, rc (etc) tags for stage release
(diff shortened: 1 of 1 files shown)


edugfilho (approved):
Perfect. Many thanks!

PR #3276, opened 2025-11-08
added lock icon due to special permissions requirement for users to participate in datatools matrix channel

D4n2021:
added lock icon, as attempting to join datatools matrix channel results in error 403 message: "MatrixError: [403] You are not invited to this room. (https://mozilla.modular.im/_matrix/client/v3/join/%23datatools%3Amozilla.org?server_name=mozilla.org&via=mozilla.org)"
and user can only click "OK" then and remains unable to send messages in that matrix channel.

### DIFF
--- a/src/components/regions/Footer.svelte
+++ b/src/components/regions/Footer.svelte
@@ -44,7 +44,7 @@
       <a href="https://www.mozilla.org/privacy/websites/#cookies">Cookies</a>
     </li>
     <li>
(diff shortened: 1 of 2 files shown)


edugfilho:
Thanks for bringing this up! @rafrombrc maybe now that GLAM is public we want to create a Matrix channel for GLAM only, or - less likely - make Data Tools public?

